PR #37, opened 2021-11-10
Ef.update connections add account

ima-hima:
Made changes to DB-API to pass `account_name` to SDK. No new tests were written.


ima-hima:
This was replaced by #59.

PR #40, opened 2021-11-17
Fir 9113: better error messages

ptiurin:
In order to ease new client sdk integration I'm improving the error messages the users are likely to get on failed connection.
This reduces the need to reach out to us when something fails and allows for faster user feedback if something that's controlled by them is incorrect (e.g. database name is mistyped or engine is not running).

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -9,7 +9,7 @@ from httpx import HTTPStatusError, RequestError, Timeout
 
 from firebolt.async_db.cursor import BaseCursor, Cursor
 from firebolt.client import DEFAULT_API_URL, AsyncClient
-from firebolt.common.exception import ConnectionClosedError, InterfaceError
+from firebolt.common.exception import ConnectionClosedError, InterfaceError, FireboltEngineError
 from firebolt.common.urls import ACCOUNT_ENGINE_URL, ENGINE_BY_NAME_URL
 from firebolt.common.util import cached_property, fix_url_schema
 
@@ -39,7 +39,9 @@ async def _resolve_engine_url(
             )
             response.raise_for_status()
             return response.json()["engine"]["endpoint"]
-        except (JSONDecodeError, RequestError, HTTPStatusError, RuntimeError) as e:
+        except HTTPStatusError as e:
+            raise FireboltEngineError(f"Firebolt engine {engine_name} does not exist")
+        except (JSONDecodeError, RequestError, RuntimeError) as e:
             raise InterfaceError(f"unable to retrieve engine endpoint: {e}")
 
 
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -11,7 +11,7 @@ from firebolt.async_db.cursor import BaseCursor, Cursor
 from firebolt.client import DEFAULT_API_URL, AsyncClient
 from firebolt.common.exception import ConnectionClosedError, InterfaceError
 from firebolt.common.urls import ACCOUNT_ENGINE_URL, ENGINE_BY_NAME_URL
-from firebolt.common.util import fix_url_schema
+from firebolt.common.util import cached_property, fix_url_schema
 
 DEFAULT_TIMEOUT_SECONDS: int = 5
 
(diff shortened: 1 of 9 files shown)


src/firebolt/async_db/connection.py
```diff
@@ -39,7 +43,16 @@ async def _resolve_engine_url(
...
+        except HTTPStatusError as e:
+            # Engine error would be 404
+            if e.response.status_code != 404:
+                raise InterfaceError(f"unable to retrieve engine endpoint: {e}")
+            # Once this is point is reached we've already authenticated with
+            # the backend so it's safe to assume the cause of the error is
+            # missing engine
+            raise FireboltEngineError(f"Firebolt engine {engine_name} does not exist")
+        except (JSONDecodeError, RequestError, RuntimeError, HTTPStatusError) as e:
+            print("Re-raised")
```
stepansergeevitch, line 55: Guess this is a debug print

src/firebolt/async_db/util.py
```diff
@@ -0,0 +1,36 @@
...
+
+from firebolt.common.urls import DATABASES_URL, ENGINES_URL
+
+
+async def is_db_available(connection, database) -> bool:
+    """Verify if the database exists"""
+    resp = await _filter_request(
+        connection, DATABASES_URL, {"filter.name_contains": database}
+    )
+    resp.raise_for_status()
```
stepansergeevitch, line 11: nit: Can be moved inside _filter_request

src/firebolt/async_db/util.py
```diff
@@ -0,0 +1,36 @@
+from httpx import Response
+
+from firebolt.common.urls import DATABASES_URL, ENGINES_URL
+
+
+async def is_db_available(connection, database) -> bool:
```
stepansergeevitch, line 6: nit: database -> database_name
please add type annotations

src/firebolt/async_db/util.py
```diff
@@ -0,0 +1,36 @@
...
+    """Verify if the database exists"""
+    resp = await _filter_request(
+        connection, DATABASES_URL, {"filter.name_contains": database}
+    )
+    resp.raise_for_status()
+    return len(resp.json()["edges"]) > 0
+
+
+async def is_engine_running(connection, engine) -> bool:
```
stepansergeevitch, line 15: nit: engine -> engine_name
Also type annotations here

src/firebolt/async_db/util.py
```diff
@@ -0,0 +1,36 @@
...
+        {
+            "filter.name_contains": engine,
+            "filter.current_status_eq": "ENGINE_STATUS_RUNNING",
+        },
+    )
+    resp.raise_for_status()
+    return len(resp.json()["edges"]) > 0
+
+
+async def _filter_request(connection, endpoint: str, filters: dict) -> Response:
```
stepansergeevitch, line 29: nit: Don't think that we need to pass whole connection here, this could only be `client` and `api_endpoint`. This way we can remove `Connection` dependency for this file
Remove connection from two previous functions if you decide to do so

src/firebolt/async_db/connection.py
```diff
@@ -158,6 +180,11 @@ async def _aclose(self) -> None:
         await self._client.aclose()
         self._is_closed = True
 
+    @cached_property
+    def _engine_name(self) -> str:
```
stepansergeevitch, line 184: Not a fan of having this field even with a private name. You think we could inline this?

stepansergeevitch (approved):
Overall LGTM. Left some minor comments, please resolve

src/firebolt/async_db/connection.py
ptiurin: Oops, missed this one. Removed.

src/firebolt/async_db/util.py
ptiurin: I agree, it can always be caught if needed.

src/firebolt/async_db/util.py
ptiurin: I'd agree if we only used `client` in `_filter_request`, but I feel like `api_endpoint` is not something whatever function or class is checking if engine/db exist should know about.

src/firebolt/async_db/connection.py
ptiurin: Fair point, this way there's no danger of using it in a place its failure would matter.

src/firebolt/async_db/util.py
ptiurin: done

src/firebolt/async_db/util.py
ptiurin: done

PR #47, opened 2021-11-18
add logging

eg-firebolt:
https://packboard.atlassian.net/browse/FIR-8571

https://packboard.atlassian.net/wiki/spaces/PACB/pages/2715779103/Logging

Current proposal is to add logging for various operations at INFO level:
* Engine - Create/Delete/Start/Stop (start includes intermediate startup statuses, so you can see progress in startup duration)
* Database - Create/Delete
* Cursor - run Query (includes query text)

Note: any of this logging can be configured/turned off. See the top of the management.ipynb notebook

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import logging
+import re
 from datetime import date, datetime
 from enum import Enum
 from functools import wraps
@@ -39,6 +41,8 @@ from firebolt.common.exception import (
 
 if TYPE_CHECKING:
     from firebolt.async_db.connection import Connection
+
+logger = logging.getLogger(__name__)
 
 ParameterType = Union[int, float, str, datetime, date, bool, Sequence]
 
@@ -219,6 +223,13 @@ class BaseCursor:
         set_parameters: Optional[Dict] = None,
     ) -> int:
         """Prepare and execute a database query. Return row count."""
+
+        # our CREATE EXTERNAL TABLE queries currently require credentials,
+        # so we will skip logging those queries.
+        # https://docs.firebolt.io/sql-reference/commands/ddl-commands#create-external-table
+        if not re.search("aws_key_id|credentials", query, flags=re.IGNORECASE):
+            logger.info(f"Running query: {query}")
+
         self._reset()
         resp = await self._do_execute_request(query, parameters, set_parameters)
         self._store_query_data(resp)
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import re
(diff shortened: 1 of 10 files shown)


src/firebolt/async_db/cursor.py
```diff
@@ -219,6 +223,13 @@ async def execute(
         set_parameters: Optional[Dict] = None,
     ) -> int:
         """Prepare and execute a database query. Return row count."""
+
+        # our CREATE EXTERNAL TABLE queries currently require credentials,
+        # so we will skip logging those queries.
+        # https://docs.firebolt.io/sql-reference/commands/ddl-commands#create-external-table
+        if not re.search("aws_key_id|credentials", query, flags=re.IGNORECASE):
+            logger.info(f"Running query: {query}")
```
ptiurin, line 231: Maybe outside of scope of this change, but later on we could log those queries as well, replacing credentials with ***. It shouldn't be a massive effort since SQL is fairly structured and those clauses will be the same unless malformed.

src/firebolt/service/binding.py
```diff
@@ -66,6 +69,9 @@ def get_database_bound_to_engine(self, engine: Engine) -> Optional[Database]:
         """Get the Database to which an engine is bound, if any."""
         try:
             binding = self.get_many(engine_id=engine.engine_id)[0]
+        except IndexError:
+            return None
+        try:
```
ptiurin, line 74: Is this intended to be a part of this change? If so, can we add tests for it?

ptiurin (commented):
I'm a little on the fence on logging the queries in INFO, but I realise there's already been discussion on this.

src/firebolt/service/binding.py
eg-firebolt: Sure I added a couple tests to cover

eg-firebolt:
> I'm a little on the fence on logging the queries in INFO, but I realise there's already been discussion on this.

This PR is the time for us to all vote and decide, so please speak your mind!

ptiurin:
I believe confirmation that sdk is working (e.g. "engine connected", "deleting database", etc.) fits really well into the INFO level, but the queries themselves are useless 90% of the time unless something is broken, from my experience.
As a next step after WARNING, INFO should provide reassurance that everything is working smoothly (if it is) and dumping queries would just clutter it. Maybe not the whole query itself but some sort of performance metric would be better logged at this level? e.g. `INFO: query run successfully in 0.01 s`
In either case, I'm not too fussed about this since the default level is WARNING and users should not get unexpected output unless they ask for it.

eg-firebolt:
> Maybe not the whole query itself but some sort of performance metric would be better logged at this level? e.g. `INFO: query run successfully in 0.01 s`

Makes sense, I updated so that DEBUG will show the whole query prior to executing, and INFO will only show the rowcount & runtime

ptiurin (approved):
LGTM, thanks for making the changes!

PR #51, opened 2021-11-23
Fix: DB error and error integration tests

ptiurin:
Fixing an issue with DB error that was not spotted before.

Adding integration tests for errors. Since EngineNotRunning error only occurs when the engine is not in a running state I propose adding another test engine that's constantly in a stopped state. This way this test tests only the error and does not involve stopping/resuming an engine that could affect other tests.

NB: this will require changing the integration tests to include a stopped engine.


tests/integration/dbapi/sync/test_errors.py
```diff
@@ -63,6 +64,25 @@ def test_engine_name_not_exists(
...
 
 
+def test_engine_stopped(
+    stopped_engine_url: str,
+    database_name: str,
+    username: str,
+    password: str,
+    api_endpoint: str,
+) -> None:
+    """Connection properly reacts to invalid engine name error"""
```
stepansergeevitch, line 74: Please update the docstring

stepansergeevitch (approved):

ptiurin:
TODO before merge: remove integration actions override

PR #52, opened 2021-11-25
Fir-9475: Allow getting cheapest instance from RM

ptiurin:
In order to be instance-name independent we want to pick the cheapest instance from the available list for our integration tests.

Limitation: There's no way to limit or filter by the type of instances (e.g. Balanced or Compute) so the cheapest instance turns out to be a compute-optimised one.


ptiurin:
As discussed with @stepansergeevitch , since this is a client-facing module we don't want to add new functionality without a proper spec. Moving this logic into the integration test workflow.

PR #58, opened 2021-11-26
enable sonar debug

stepansergeevitch:


eg-firebolt:
Can we add a wiki page or something that describes what SonarCloud is and why we picked it? And some guidance around what do do if certain detections occur?

ima-hima:
> Can we add a wiki page or something that describes what SonarCloud is and why we picked it? And some guidance around what to do if certain detections occur?

Yes, please!

stepansergeevitch:
@eg-firebolt @ima-hima Full CI documentation for all processes should be ready soon, I will notify you.

PR #60, opened 2021-11-30
ci: FIR-9440 automated release workflow

ptiurin:
Automating version increase and release to github and pypi.

Before merge we'll need to add pypi credentials to github secrets.
Token will be added in username/password secrets according to https://test.pypi.org/help/#apitoken


.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
...
+        git tag ${{ steps.tag_generation.outputs.new_tag }}
+        git push origin ${{ steps.tag_generation.outputs.new_tag }}
+        
+    - name: Publish to pypi
+      run: |
+        python3 -m build
+        python3 -m twine upload dist/*
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
```
eg-firebolt, line 65: new line at end of file?

ci/generate_version_tag.py
```diff
@@ -0,0 +1,48 @@
...
+        else:
+            # Enhancements bump minor version while anything else is considered a patch
+            enhancements = [c for c in changes.split("\n") if c.startswith("enh:")]
+            new_ver = ver.bump_minor() if enhancements else ver.bump_patch()
+    print(new_ver)
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(
+        description="Determine the next tag based on change tags"
```
eg-firebolt, line 36: Some additional context about this script would be helpful either here or in a docstring at the top of the file.

ci/generate_version_tag.py
```diff
@@ -0,0 +1,48 @@
+import argparse
+
+import semver
+
+
+def main(changes, old_tag, prerelease_tag, major_release):
```
eg-firebolt, line 6: Could we add typehints and a usual docstring that describes each parameter?

ptiurin: Done.

ci/generate_version_tag.py
ptiurin: Fair point, added a docstring.

ci/generate_version_tag.py
```diff
@@ -0,0 +1,76 @@
...
+    ver = semver.VersionInfo.parse(old_tag)
+    if major_release:
+        new_ver = ver.bump_major()
+    else:
+        if not changes:
+            raise Exception("No changes since the last release!")
+        # Breaking changes are of form <tag>!: <description>
+        # To avoid automatically publishing major version they need to be
+        # explicitly specified
+        breaking = [c for c in changes.split("\n") if "!:" in c]
```
stepansergeevitch, line 43: We need to make sure that no one has `!:` in their PR name accidentally (i guess that's pretty unlikely)
Other solution could be to do `c.split(':', 1)[0].endswith('!')`

.github/workflows/integration-tests.yml
```diff
@@ -1,6 +1,7 @@
 name: Run integration tests
 on:
   workflow_dispatch:
+  workflow_call:
```
stepansergeevitch, line 4: Are we reusing this workflow somewhere?

stepansergeevitch (approved):

.github/workflows/integration-tests.yml
ptiurin: Good spot, I was meant to add integration test check before release, but removed it for testing.

ci/generate_version_tag.py
ptiurin: Yeah, doesn't look like it's spotted by the PR title checker so better be safe.

PR #63, opened 2021-12-01
ci: Fir 9443 create nightly jobs

stepansergeevitch:
Created a nightly job for SDK, runs on windows, macos and linux. Uses python 3.7, 3.8, 3.9, 3.10.
Runs code check, unit and integration tests

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -7,6 +7,7 @@ jobs:
   code-check:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false # finish all jobs even if one fails
       matrix: 
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: ['3.7', '3.8', '3.9', '3.10']
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -35,6 +35,6 @@ jobs:
         run: |
           pytest tests/unit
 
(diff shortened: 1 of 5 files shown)


eg-firebolt:
Does the nightly job run against the `main` branch, a specific release, or something else?

stepansergeevitch:
@eg-firebolt Sorry, didn't see your comment. The scheduled nightly job always runs against main branch. Although, it can be run manually against any branch

stepansergeevitch:
The job is passing now: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/1563393534

.github/workflows/nightly.yml
```diff
@@ -5,11 +5,38 @@ on:
     - cron: '0 3 * * *' # 3 am UTC every day
 jobs:
   code-check:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check out code
```
ptiurin, line 10: Did you mean to reuse here .github/workflows/code-check.yml ?

ptiurin (approved):
Apart from that lgtm

.github/workflows/nightly.yml
stepansergeevitch: I don't like how github let's you reuse jobs because you need to specify specific branch/commit when reusing, and you can't just say use the same commit as the calling job. This makes it harder to test jobs on specific branches. That's why I think it's better to avoid reusing jobs when possible

PR #68, opened 2021-12-02
Added Read the Docs config

marknoack:
added 

- readthedocs.yaml

- requirements.txt

- modified /docsrc files for better docstring imports


.readthedocs.yaml
```diff
@@ -0,0 +1,6 @@
+sphinx:
+  configuration: docsrc/conf.py
+
+python:
+  install:
+    - requirements: docs/requirements.txt
```
eg-firebolt, line 6: add newline at end of file

docs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pydantic
+httpx
+aiorwlock
+async_property
+readerwriterlock
```
eg-firebolt, line 5: What does this file do for us? Why can't we use the requirements listed in setup.cfg?

docsrc/conf.py
```diff
@@ -13,8 +13,17 @@
...
+sys.path += [
+	os.path.abspath("../src"), 
+	os.path.abspath("../src/firebolt/"), 
+	os.path.abspath("../src/firebolt/async_db"), 
+	os.path.abspath("../src/firebolt/client"), 
+	os.path.abspath("../src/firebolt/common"),
+	os.path.abspath("../src/firebolt/db"),
+	os.path.abspath("../src/firebolt/model"), 
+	os.path.abspath("../src/firebolt/service")
+	]
```
eg-firebolt, line 25: This may be brittle if/when we refactor modules. For example, I plan to restructure modules once some of the current PRs are done: https://packboard.atlassian.net/browse/FIR-9390

Why do we need `"../src"` and also subpaths like `"../src/firebolt/"` etc?

src/firebolt/service/types.py
```diff
@@ -37,74 +37,74 @@ class EngineStatus(Enum):
 
     ENGINE_STATUS_UNSPECIFIED = "ENGINE_STATUS_UNSPECIFIED"
 
-    # Logical record is created, however underlying infrastructure is not initialized.
-    # In other words this means that engine is stopped.
+    ''' Logical record is created, however underlying infrastructure is not initialized.
```
eg-firebolt, line 40: These should probably be double quotes everywhere, since that is compliant with our linter, black: https://github.com/psf/black

docsrc/conf.py
marknoack: The "../src/firebolt/" line can probably be removed, but unfortunately listing out each and every module folder was the only way ReadtheDocs would import everything

docs/requirements.txt
marknoack: The requirements listed here are duplicative of what's in the setup.cfg file, however this is how ReadtheDocs encourages users to declare them : https://docs.readthedocs.io/en/stable/config-file/v2.html

PR #70, opened 2021-12-03
fix: add custom loop for connection __del__

stepansergeevitch:
Apparently, I've finally managed to investigate the issue with `RuntimeWarning: coroutine 'BaseConnection._aclose' was never awaited` in tests.
It seems like `connection.__del__` function (which inside calls asynchronous `_aclose` in global event loop) sometimes gets called after the event loop is closed (because it's called by the garbage collector in background). To fix this, added a custom loop for connection to use it for closing. I assume this issue could happen in real environments too.

### DIFF
--- a/src/firebolt/db/connection.py
+++ b/src/firebolt/db/connection.py
@@ -59,8 +59,9 @@ class Connection(AsyncBaseConnection):
     @wraps(AsyncBaseConnection._aclose)
     def close(self) -> None:
         with self._closing_lock.gen_wlock():
-            self._loop.run_until_complete(self._aclose())
-            self._loop.close()
+            if not self.closed:
+                self._loop.run_until_complete(self._aclose())
+                self._loop.close()
 
     # Context manager support
     def __enter__(self) -> Connection:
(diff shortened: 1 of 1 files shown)


src/firebolt/db/connection.py
```diff
@@ -57,7 +59,7 @@ def cursor(self) -> Cursor:
     @wraps(AsyncBaseConnection._aclose)
     def close(self) -> None:
         with self._closing_lock.gen_wlock():
-            return async_to_sync(super()._aclose)()
+            return self._loop.run_until_complete(self._aclose())
```
ptiurin, line 62: Would using https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop be beneficial in here since this way if a loop is still running we don't create a new one?

stepansergeevitch: Well, I actually don't think that creating an additional loop rather than reusing an existing one would be much slower in this case. I think that here it's more just a matter of organization since we only run one task at most in parallel

stepansergeevitch: Also, the biggest point here is to be able to control when the loop is being closed

ptiurin: I think we run one task at a time anyway, but I agree with the second point, better to be explicit when closing this loop.

ptiurin (approved):

PR #73, opened 2021-12-08
test: temporary disable long running integration tests

stepansergeevitch:
Temporary disabled long running integration tests until we figure out what's wrong with it.
This tests was doing insert and than select which checked inserted data. We still run a single insert query now but do not check the data.


ptiurin (approved):
LGTM, better have some tests than none. Let's file a ticket to investigate the hanging jobs.

stepansergeevitch:
Investigation ticket: https://packboard.atlassian.net/browse/FIR-9825

PR #76, opened 2021-12-09
feat: implement Connection.commit, DataError on empty data fetch

### LINKED ISSUE #74: Missing commit() Connection method
```
.commit()
...

Database modules that do not support transactions should implement this method with void functionality.
~ PEP-249 (https://www.python.org/dev/peps/pep-0249/#commit)
```
I know it's just should, but it would be nice to add it and mention the lack of transactions in Firebolt engine.

stepansergeevitch:
Improvements for PEP-249 conformance
- Added `connection.commit` method with void functionality
- Raise `DataError` when no data was returned from query

closes #74

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -191,7 +191,7 @@ class BaseConnection:
         except ValueError:
             pass
 
-    def commit(self):
+    def commit(self) -> None:
         """Does nothing since Firebolt doesn't have transactions"""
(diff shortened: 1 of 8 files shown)


ptiurin (approved):
LGTM. Nitpicking, but let's rename the PR title/commit to something more descriptive, like `implement commit and data error` or something along the lines.

eg-firebolt (approved):

PR #79, opened 2021-12-13
ci: Reuse code-check workflow in nightly, split python-app workflow

stepansergeevitch:
Some minor ci improvements before documenting:
- Reuse code-check in nightly
- Splitted python-app workflow into separate pieces


ptiurin (commented):
I like the idea of splitting unit tests and code check out. Though maybe we should have a generic CI workflow that unifies those, for clarity. This way we'd have a single workflow with several checks on PR e.g.
```
CI/Code style -- SUCCESS
CI/Unit tests -- SUCCESS
CI/PR Title -- FAIL
```
Similar to nightly.yml
```
jobs:
  code-style:
    uses: .../code-style.yml
  unit-tests:
    uses: .../unit-tests.yml
  pr-check:
    uses: jef/conventional-commits-pr-action@v1.0.0. (since there's no point in moving it out)
```

ptiurin (approved):
LGTM

PR #86, opened 2021-12-16
feat: Add account_name to DB-API

ima-hima:
Allow specific accounts to be used when interfacing with API. Already exists in SDK.


ima-hima:
Used incorrect account.

src/firebolt/async_db/connection.py
```diff
@@ -130,6 +141,7 @@ class BaseConnection:
         "_cursors",
         "database",
         "engine_url",
+        "account_name",
```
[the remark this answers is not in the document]
ericf-firebolt: I don't think we should pull it from here. It's not actually being stored, but being used to initialized an `AsyncClient` object. `AsyncClient` needs the `account_name` to retrieve the `account_id`, which is in turn needed for several engine-related interactions with the API. As example is on l. 36, inside `_resolve_engine_url()`. Were `account_name` to be set as `None`, it's possible/likely that the incorrect `account_id` would be retrieved, meaning that the wrong `engine_url` would be created.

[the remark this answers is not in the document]
stepansergeevitch: `account_id` is never used in `Connection`
It's ok to pass it to `connect` function, but no need to pass it to `Connection`

src/firebolt/client/client.py
```diff
@@ -54,7 +56,15 @@ class Client(FireboltClientMixin, HttpxClient):
 
     @cached_property
     def account_id(self) -> str:
-        return self.get(url=ACCOUNT_URL).json()["account"]["id"]
+        if self.account_name is not None:
+            response = self.get(
+                url=ACCOUNT_BY_NAME_URL, params={"account_name": self.account_name}
+            )
+            if response.status_code != 200:
```
stepansergeevitch, line 63: Let's raise AccountError only on 404 code.
In all other cases let's propagate the error further.
```
if response.status_code == httpx.codes.NOT_FOUND:
    raise AccountError(self.account_name)
# process all other status codes
response.raise_for_status()
```

src/firebolt/common/util.py
```diff
@@ -1,7 +1,8 @@
-from asyncio import get_event_loop, new_event_loop, set_event_loop
-from functools import lru_cache, wraps
+from functools import lru_cache, partial, wraps
 from typing import TYPE_CHECKING, Any, Callable, Type, TypeVar
 
+from trio import run as sync_run
```
stepansergeevitch, line 4: We still use asyncio for async stuff. Please rebase on the latest master

src/firebolt/common/exception.py
```diff
@@ -29,6 +29,14 @@ class FireboltDatabaseError(FireboltError):
     pass
 
 
+class AccountError(FireboltError):
```
stepansergeevitch, line 32: `AccountError` -> `AccountNotFoundError` or something more verbose

src/firebolt/common/util.py
ericf-firebolt: Oops! I had merged into the wrong local branch.

src/firebolt/common/exception.py
ericf-firebolt: Done.

src/firebolt/async_db/connection.py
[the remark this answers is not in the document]
ericf-firebolt: Ah. I see what you're saying. How's this?

This isn't going to cause issues with Cursors that only use default account_ids, right?

src/firebolt/client/client.py
ericf-firebolt: See query re Cursors with default account_id values above. Otherwise done.

setup.cfg
```diff
@@ -24,11 +24,10 @@ project_urls =
 packages = find:
 install_requires =
     aiorwlock==1.1.0
-    async-property==0.2.1
     httpx[http2]==0.19.0
     pydantic[dotenv]==1.8.2
     readerwriterlock==1.0.9
-    syncer==1.3.0
+    trio==0.19.0
```
stepansergeevitch, line 30: No need to it since we don't use it

tests/unit/async_db/test_connection.py
```diff
@@ -55,11 +55,14 @@ async def test_cursor_initialized(
...
     query_callback: Callable,
     query_url: str,
+    account_id_callback: Callable,
+    account_id_url: str,
     python_query_data: List[List[ColType]],
 ) -> None:
     """Connection initialised it's cursors propperly"""
     httpx_mock.add_callback(auth_callback, url=auth_url)
     httpx_mock.add_callback(query_callback, url=query_url)
+    # httpx_mock.add_callback(account_id_callback, url=account_id_url)
```
stepansergeevitch, line 65: remove comment

tests/unit/conftest.py
```diff
@@ -124,19 +126,30 @@ def db_name() -> str:
 
 @pytest.fixture
 def account_id_url(settings: Settings) -> str:
-    return f"https://{settings.server}{ACCOUNT_URL}"
+    if settings.account_name is None:
```
stepansergeevitch, line 129: Could account_name ever be None in our tests?
I think we require ACCOUNT_NAME env variable to be provided.
So I think we can leave only second branch

stepansergeevitch (commented):
Overall LGTM, please resolve new comments

setup.cfg
```diff
@@ -49,6 +48,7 @@ dev =
     pytest-cov==3.0.0
     pytest-httpx==0.13.0
     pytest-mock==3.6.1
+    trio-typing==0.7.0
```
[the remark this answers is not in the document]
ericf-firebolt: Here too.

src/firebolt/client/client.py
```diff
@@ -70,6 +83,15 @@ class AsyncClient(FireboltClientMixin, HttpxAsyncClient):
     + (HttpxAsyncClient.__doc__ or "")
     """
 
-    @async_cached_property
```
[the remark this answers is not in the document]
ericf-firebolt: Yep. This was a rebase failure.

tests/unit/async_db/test_connection.py
ericf-firebolt: Done.

tests/unit/conftest.py
ericf-firebolt: It can be `None`. It's actually set to `None` on l. 98 of `conftest.py` as a default and fails all over the place when I remove the first branch.

setup.cfg
```diff
@@ -24,11 +24,9 @@ project_urls =
 packages = find:
 install_requires =
     aiorwlock==1.1.0
-    async-property==0.2.1
```
stepansergeevitch, line 27: I think this one is required

tests/unit/async_db/test_connection.py
ericf-firebolt: Done.

tests/unit/conftest.py
ericf-firebolt: I also changed this logic to deal with the possibility of `settings.account+name` being an empty string.

setup.cfg
ericf-firebolt: Done.

stepansergeevitch (approved):

PR #88, opened 2021-12-27
feat: migrate from asyncio to trio

stepansergeevitch:
We need to migrate from using asyncio as async backend to trio.
It seems like asyncio has a bug, which causes tcp transport to raise "Event loop closed" error on close in some cases.
Trio, on the other hand, seems not to have this issue

### DIFF
--- a/src/firebolt/client/client.py
+++ b/src/firebolt/client/client.py
@@ -1,6 +1,5 @@
 from typing import Any, Optional
 
-from async_property import async_cached_property  # type: ignore
 from httpx import AsyncClient as HttpxAsyncClient
 from httpx import Client as HttpxClient
 from httpx import _types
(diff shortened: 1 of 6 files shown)


src/firebolt/client/client.py
```diff
@@ -70,6 +69,5 @@ class AsyncClient(FireboltClientMixin, HttpxAsyncClient):
     + (HttpxAsyncClient.__doc__ or "")
     """
 
-    @async_cached_property
```
ptiurin, line 73: Any reason `account_id` is no longer a property?

ptiurin (commented):
As far as migration goes, this seems to be pretty seamless so lgtm.
However, I'm wary about moving to a different async library. Asyncio is an industry standard with many usecases tested daily, while trio looks like a fairly recent one. We'll be losing a lot in terms of support - asyncio has 5000+ questions on Stack Overflow, while trio has 59.
Is "Event loop closed" error a known issue in asyncio? If not, can we create a reproducer and raise an issue on their repo to see what the maintainers say?

PR #90, opened 2021-12-30
feat: Scan code for security issues

ptiurin:
Adding security scan GH action. It's supposed to be run manually on a PR, at the same time as the integration tests. It also runs automatically on merge to master.
Using pipenv to build the project in order for Fossa to pick deep dependencies according to the [docs](https://github.com/fossas/fossa-cli/blob/master/docs/references/strategies/languages/python/python.md).

### DIFF
--- a/.github/workflows/security-scan.yml
+++ b/.github/workflows/security-scan.yml
@@ -13,36 +13,5 @@ jobs:
       - name: "Checkout Code"
         uses: actions/checkout@v2
 
-      - name: "Fossa branch fix"
-        if: ${{ github.event_name == 'pull_request' }}
-        run: |
-          git fetch
-          git branch --track ${GITHUB_HEAD_REF##*/} origin/${GITHUB_HEAD_REF##*/}
-          git checkout ${GITHUB_HEAD_REF##*/}
-
-      - name: "Build"
-        run: |
-          python3 -m pip install pipenv
-          python3 -m pipenv install -e .
-
-      - name: "Run FOSSA Scan"
-        uses: fossas/fossa-action@v1
-        with:
-          api-key: ${{secrets.fossaApiKey}}
-
-  sonar-scan:
-    runs-on: ubuntu-latest
-    steps:
-      - name: "Checkout Code"
-        uses: actions/checkout@v2
-
-      - name: SonarCloud Scan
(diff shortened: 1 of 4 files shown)


.github/workflows/security-scan.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout Code"
+        uses: actions/checkout@v2
+
+      - name: "Security Scan"
+        uses: firebolt-db/action-security-scan@main
+        with:
+          github-key: ${{ secrets.GITHUB_TOKEN }}
+          fossa-key: ${{ secrets.fossaApiKey }}
```
stepansergeevitch, line 20: Is it possible to have this secret in upper snake case?
Like `secrets.FOSSA_API_KEY`

stepansergeevitch (approved):
Overall LGTM.
One issue is that currently sonar cloud seems to fail because of no coverage.
I assume that we would not provide coverage soon
Is it possible to disable the coverage check so it passes?

stepansergeevitch:
Also, it seems like FOSSA fails with an exception

.github/workflows/security-scan.yml
ptiurin: Yeah, once we have the org-wide secrets they'll be upper-case and I will remove this one.

ptiurin:
> Is it possible to disable the coverage check so it passes?

Unfortunately no, sonarqube/cloud does not let us disable the check. It's annoying, but we might need to ignore this check ourselves or re-visit a way to provide coverage information.
We've seen it before being wonky, failing or not failing for no particular reason.

ptiurin:
> Also, it seems like FOSSA fails with an exception

Followed up on IM: Fossa decoration is not working for some reason. I've added the next recommended thing - fossa-test that fails the PR if issues are found. This is not the build itself, it's FOSSA finding problems (which we'll address shortly).

ptiurin:
In fact, both issues Fossa has found are false-positives. I didn't resolve them in order to provide an indication what will happen if such issues are found.

.github/workflows/security-scan.yml
ptiurin: Fixed so there's no need to change this later.

PR #94, opened 2022-01-10
feat: Implement prepared statements

stepansergeevitch:
Implemented prepared statements based on the design doc https://docs.google.com/document/d/1oWDgbSjI1NvIJJJlgb5Yqrsqw4vIQgkEMjn1XXG1NWM/edit


src/firebolt/async_db/_types.py
```diff
@@ -190,6 +191,64 @@ def parse_value(
...
+    elif isinstance(value, datetime):
+        if value.tzinfo is not None:
+            value = value.astimezone(timezone.utc)
+        return f"'{value.strftime('%Y-%m-%d %H:%M:%S')}'"
+    elif isinstance(value, date):
+        return f"'{value.isoformat()}'"
+    if value is None:
+        return "NULL"
+    elif isinstance(value, Sequence):
+        return f"[{', '.join(format_value(it) for it in value)}]"
```
ptiurin, line 218: Would sequence of strings need to be quoted? e.g. `['a', 'b', 'c']` ?

src/firebolt/async_db/_types.py
```diff
@@ -190,6 +191,64 @@ def parse_value(
...
+
+    raise DataError(f"unsupported parameter type {type(value)}")
+
+
+def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
```
ptiurin, line 223: Just thinking, could this be done with a `re.sub()`? So there's less re-inventing the wheel, as I feel like the problem of replacing the placeholders is not new.

src/firebolt/async_db/cursor.py
```diff
@@ -271,6 +270,10 @@ async def executemany(
         Prepare and execute a database query against all parameter
         sequences provided. Return last query row count.
         """
+
+        if len(parameters_seq) > 1:
+            raise NotSupportedError("multi-statement queries are not supported")
```
ptiurin, line 275: Should this be "Parametrised multi-statement queries are not supported"?

tests/unit/async_db/test_typing_format.py
```diff
@@ -0,0 +1,72 @@
...
+from pytest import raises
+
+from firebolt.async_db import DataError
+from firebolt.async_db._types import format_sql, format_value
+
+
+def test_format_value() -> None:
+    test_cases = (
+        # Strings
+        ("abcd", "'abcd'"),
```
ptiurin, line 12: Any advantage over `@pytest.mark.parametrize` here?

src/firebolt/async_db/_types.py
```diff
@@ -190,6 +191,64 @@ def parse_value(
...
+    backtick_cnt, last_position, parameter_idx = 0, 0, 0
+    parts = []
+    for i, ch in enumerate(query):
+        if ch == "?":
```
ptiurin, line 227: This would replace '?' in any part of the query. Do we want to support things like `INSERT INTO ? VALUES(1, 2)` or even allowing dynamically replacing the keywords like INSERT with '?'?

src/firebolt/async_db/_types.py
stepansergeevitch: Each string will be quoted inside `format_value` function

src/firebolt/async_db/cursor.py
stepansergeevitch: Yeah, we can use that

tests/unit/async_db/test_typing_format.py
stepansergeevitch: Good point, added it

src/firebolt/async_db/_types.py
stepansergeevitch: Making it context dependent would make the implementation much more complicated.
I guess allowing users to do so would harm in any way. What do you think?

ptiurin (approved):
LGTM for this approach. Let's see what the consensus is for the SQL parsing suggestion.

tests/unit/async_db/test_typing_format.py
```diff
@@ -0,0 +1,76 @@
...
+        (False, "0"),
+        # Date, datetime
+        (date(2022, 1, 10), "'2022-01-10'"),
+        (datetime(2022, 1, 10, 1, 1, 1), "'2022-01-10 01:01:01'"),
+        (
+            datetime(2022, 1, 10, 1, 1, 1, tzinfo=timezone(timedelta(hours=1))),
+            "'2022-01-10 00:01:01'",
+        ),
+        # List, tuple
+        ([1, 2, 3], "[1, 2, 3]"),
```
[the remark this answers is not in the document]
stepansergeevitch: Sure, will add

src/firebolt/async_db/_types.py
```diff
@@ -190,6 +192,67 @@ def parse_value(
...
+        return f"[{', '.join(format_value(it) for it in value)}]"
+
+    raise DataError(f"unsupported parameter type {type(value)}")
+
+
+# Find an escaped(\?) or plain(?) question mark
+sql_param_re = compile(r"(\\\?)|\?")
+
+
+def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
```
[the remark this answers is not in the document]
stepansergeevitch: Yeah, that's a good point.
I think what we can do in this case is remove all comments from the query before substituting parameters

ptiurin (approved):

yuryfirebolt (changes requested):

src/firebolt/async_db/_types.py
```diff
@@ -190,6 +195,69 @@ def parse_value(
...
+        if isinstance(token, TokenList):
+            # Process all children tokens
+            token.tokens = [process_token(t) for t in token.tokens]
+        return token
+
+    parsed = parse_sql(query)
+    if len(parsed) > 1:
+        raise NotSupportedError("Multi-statement queries are not supported")
+
+    formatted_sql = str(process_token(parsed[0]))
```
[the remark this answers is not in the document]
stepansergeevitch: Good point. Fixed and added test

yuryfirebolt (approved):

PR #99, opened 2022-01-13
test: fix integration tests

stepansergeevitch:
Add ACCOUNT_NAME env variable to test runs


yuryfirebolt (approved):

ptiurin (approved):
Looks good, but integration tests are failing with a timeout so we can't verify this works.

PR #103, opened 2022-01-18
fix: hanging on engine start, if engine starting failed

yuryfirebolt:


ptiurin (commented):
Looks good. Any tests we can add to test this?

yuryfirebolt:
> Looks good. Any tests we can add to test this?

Probably not with the integration tests, only with unit-tests and mock. So far, there are no unit-test for start engine, so I was not able to extend it.

stepansergeevitch (approved):

PR #105, opened 2022-01-19
fix: databases.get_many function, add integration test for it

yuryfirebolt:


tests/integration/resource_manager/test_engine.py
```diff
@@ -47,3 +47,17 @@ def test_copy_engine():
...
+
+    # get all databases, at least one should be returned
+    databases = rm.databases.get_many()
+    assert len(databases) > 0
+    assert database_name in {db.name for db in databases}
+
+    # get all databases, with name_contains
+    databases = rm.databases.get_many(name_contains=database_name)
+    assert len(databases) > 0
+    assert database_name in {db.name for db in databases}
```
stepansergeevitch, line 63: Please also test passing `attached_engine_name_eq` and `attached_engine_name_contains` since we know it has engines.
Also, do you think we can test `order_by`?
We can just see how many databases there are from the first call, and if it's more that one we also try passing `order_by`

stepansergeevitch (commented):
Looks good. Left some comment regarding tests

tests/integration/resource_manager/test_engine.py
yuryfirebolt: Added tests for `attached_engine_name_eq` and `attached_engine_name_contains`. However, I would not do additional tests on `order_by`, since it looks likes we are re-implementing order_by functionality in the test

stepansergeevitch (approved):

tests/integration/resource_manager/test_engine.py
ptiurin: We can still do some basic checks on `order_by` without re-implementing the functionality. e.g. checking that the parameter is propagated correctly and is added to the http request.

PR #108, opened 2022-01-20
test: Make pytest generate coverage report and sonar read it

yuryfirebolt:
The pipeline fails on the branch, because it depends on the changes, that should be in main

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,10 +15,10 @@ jobs:
   code-checkers:
     uses: firebolt-db/firebolt-python-sdk/.github/workflows/code-check.yml@main
   unit-tests:
-    uses: firebolt-db/firebolt-python-sdk/.github/workflows/unit-tests.yml@main
+    uses: firebolt-db/firebolt-python-sdk/.github/workflows/unit-tests.yml@add-coverage-report
   security-scan:
     needs: [unit-tests]
-    uses: firebolt-db/firebolt-python-sdk/.github/workflows/security-scan.yml@main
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request.yml
```diff
@@ -16,3 +16,10 @@ jobs:
     uses: firebolt-db/firebolt-python-sdk/.github/workflows/code-check.yml@main
   unit-tests:
     uses: firebolt-db/firebolt-python-sdk/.github/workflows/unit-tests.yml@main
+  security-scan:
+    needs: [unit-tests]
+    uses: firebolt-db/firebolt-python-sdk/.github/workflows/security-scan.yml@main
```
ptiurin, line 21: you can use `security-scan.yml@<your_branch_name>` to test

ptiurin (commented):
I don't like the coupling between security-scan and the rest of CI. Ideally, security-scan should be a plug and play as we expose it through the templates: https://github.com/firebolt-db/.github/blob/main/workflow-templates/security-scan-org.yml (setup doc: https://packboard.atlassian.net/wiki/spaces/PACB/pages/2813231305/New+Project+CI+setup).
I'm thinking it should stay a stand-alone workflow, with a step that calculates the coverage. This step would be a placeholder for projects to add their own coverage calculations (since we have other language projects apart from Python).

.github/workflows/pull-request.yml
yuryfirebolt: yes, did it for testing purposes in a separate branch, now want to push it to main

ptiurin (approved):
As discussed on the daily, this is a necessary step. We'll need to modify the templates after this change is confirmed to include this new dependency.

PR #110, opened 2022-01-20
feat: Fir 8425 implement multi statement funct

stepansergeevitch:
Implemented multi-statement queries functionality based on the design
https://docs.google.com/document/d/13O5jHFjnvRrNSvzBXNohYYz4ChI1jhi_7xyIwd1_7Xs/edit
- Added `nextset` method from PEP249 to be able to move to next result set
- Set `cursor.state` to new `CursorState.ERROR` value if error occured during query execution. This allows to figure out if all queries have passed in case of multi-statement, and is more verbose in my opinion

### DIFF
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -265,7 +265,7 @@ def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
 
 
 def split_format_sql(
-    query: str, parameters: Sequence(Sequence[ParameterType])
+    query: str, parameters: Sequence[Sequence[ParameterType]]
 ) -> List[str]:
     statements = parse_sql(query)
     if not statements:
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -245,7 +245,8 @@ def format_statement(statement: Statement, parameters: Sequence[ParameterType]) 
             return Token(TokenType.Text, formatted)
         if isinstance(token, TokenList):
             # Process all children tokens
-            token.tokens = [process_token(t) for t in token.tokens]
+
+            return TokenList([process_token(t) for t in token.tokens])
         return token
 
     formatted_sql = str(process_token(statement)).rstrip(";")
@@ -263,7 +264,9 @@ def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
     return format_statement(parse_sql(query)[0], parameters)
 
 
-def split_format_sql(query: str, parameters: Sequence[ParameterType]) -> List[str]:
+def split_format_sql(
+    query: str, parameters: Sequence(Sequence[ParameterType])
+) -> List[str]:
     statements = parse_sql(query)
     if not statements:
         return [query]
@@ -273,5 +276,5 @@ def split_format_sql(query: str, parameters: Sequence[ParameterType]) -> List[st
             raise NotSupportedError(
                 "formatting multistatement queries is not supported"
             )
-        return [format_statement(statements[0], parameters)]
+        return [format_statement(statements[0], paramset) for paramset in parameters]
     return [str(st).strip().rstrip(";") for st in statements]
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -226,8 +226,7 @@ def format_value(value: ParameterType) -> str:
 
 def format_statement(statement: Statement, parameters: Sequence[ParameterType]) -> str:
     """
-    Substitute placeholders in queries with provided values.
-    '?' symbol is used as a placeholder. Using '\\?' would result in a plain '?'
+    Substitute placeholders in a sqlparse statement with provided values.
     """
     idx = 0
 
@@ -260,13 +259,14 @@ def format_statement(statement: Statement, parameters: Sequence[ParameterType]) 
     return formatted_sql
 
 
-def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
-    return format_statement(parse_sql(query)[0], parameters)
-
-
 def split_format_sql(
     query: str, parameters: Sequence[Sequence[ParameterType]]
 ) -> List[str]:
+    """
+    Split a query into separate statement, and format it with parameters
+    if it's a single statement
+    Trying to format a multi-statement query would result in NotSupportedError
+    """
     statements = parse_sql(query)
     if not statements:
         return [query]
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -3,10 +3,11 @@ from __future__ import annotations
 from collections import namedtuple
 from datetime import date, datetime, timezone
 from enum import Enum
-from typing import Sequence, Union
+from typing import List, Sequence, Union
 
 from sqlparse import parse as parse_sql  # type: ignore
-from sqlparse.sql import Token, TokenList  # type: ignore
+from sqlparse import split
+from sqlparse.sql import Statement, Token, TokenList  # type: ignore
 from sqlparse.tokens import Token as TokenType  # type: ignore
 
 try:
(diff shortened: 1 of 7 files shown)


src/firebolt/async_db/_types.py
```diff
@@ -263,3 +258,23 @@ def process_token(token: Token) -> Token:
         )
 
     return formatted_sql
+
+
+def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
+    return format_statement(parse_sql(query)[0], parameters)
+
```
ptiurin, line 265: Is this function used anywhere?

src/firebolt/async_db/_types.py
```diff
@@ -263,3 +258,23 @@ def process_token(token: Token) -> Token:
...
+
+def split_format_sql(
+    query: str, parameters: Sequence[Sequence[ParameterType]]
```
ptiurin, line 268: Could you add a short docstring for this function? Despite not being public it would be useful to have down the line.

src/firebolt/async_db/cursor.py
```diff
@@ -164,24 +173,59 @@ def close(self) -> None:
...
+                # Parse data during fetch
+                rows = query_data["data"]
+                row_set = (rowcount, descriptions, rows)
+            except (KeyError, JSONDecodeError) as err:
+                raise DataError(f"Invalid query data format: {str(err)}")
+
+        self._row_sets.append(row_set)
+        if self._next_set_idx == 0:
+            # Populate values for first set
+            self._pop_next_set()
```
ptiurin, line 204: Does this mean if we do `cursor.execute(query1; query2)` -> `cursor.nextset()`(popping one set) -> `cursor.execute(query3)` -> `cursor.nextset()` then the result we're holding will be the result of query2 not query3? Is this expected behaviour?

src/firebolt/async_db/cursor.py
```diff
@@ -213,29 +257,52 @@ def _reset(self) -> None:
...
+                    f" {time.time() - start_time} seconds"
+                )
+
+            self._state = CursorState.DONE
 
-        await self._raise_if_error(resp)
-        return resp
+        except Exception:
+            self._state = CursorState.ERROR
+            raise
```
ptiurin, line 305: Do we check anywhere for CursorState.ERROR? Or is this purely for user information?

src/firebolt/async_db/cursor.py
```diff
@@ -245,21 +312,9 @@ async def execute(
...
-            logger.debug(f"Running query: {query}")
-
-        self._reset()
-        resp = await self._do_execute_request(query, parameters, set_parameters)
-        self._store_query_data(resp)
-        self._state = CursorState.DONE
-        logger.info(
-            f"Query fetched {self.rowcount} rows in {time.time() - start_time} seconds"
-        )
+        params_list = [parameters] if parameters else []
```
ptiurin, line 316: I'm not entirely understanding this line. As I understand the user would call `execute("INSERT ... VALUES (?, ?)", [(1,2), (3,4), (5,6)])` is that not what we want to pass to the _do_execute_request?

src/firebolt/async_db/_types.py
stepansergeevitch: Oh, no. Should remove it
It was used temporary

src/firebolt/async_db/_types.py
stepansergeevitch: Yes, sure

src/firebolt/async_db/cursor.py
stepansergeevitch: If you do another execute, the `_row_sets` is being cleared, so you won't able to read results from previous execute.

src/firebolt/async_db/cursor.py
stepansergeevitch: Yes, it's only for users

src/firebolt/async_db/cursor.py
yuryfirebolt: I think `reset` is called on execute, which would make the `nextset` return the result only from the last execution

src/firebolt/async_db/cursor.py
stepansergeevitch: No, users can only pass a list of parameters to executemany.
In case of execute, only a single set of parameters is passed.
So you can only do `execute("INSERT ... VALUE (?, ?)", (1, 2))`

src/firebolt/async_db/cursor.py
```diff
@@ -164,24 +173,59 @@ def close(self) -> None:
...
+        return self._pop_next_set()
+
+    def _pop_next_set(self) -> Optional[bool]:
+        """
+        Same functionality as .nextset, but doesn't check that query has been executed.
+        """
+        if self._next_set_idx >= len(self._row_sets):
+            return None
+        cur_set = self._row_sets[self._next_set_idx]
+        self._rowcount = cur_set[0]
```
yuryfirebolt, line 224: Can we use unpacking here: `self._rowcount, self._descriptions, self._rows = self._row_sets[self._next_set_idx]`?

src/firebolt/async_db/cursor.py
stepansergeevitch: From PEP249
execute: `Parameters may be provided as sequence or mapping and will be bound to variables in the operation`
executemany: `Prepare a database operation (query or command) and then execute it against all parameter sequences or mappings found in the sequence seq_of_parameters.`

yuryfirebolt (changes requested):

src/firebolt/async_db/_types.py
```diff
@@ -263,3 +258,23 @@ def process_token(token: Token) -> Token:
         )
 
     return formatted_sql
+
+
+def format_sql(query: str, parameters: Sequence[ParameterType]) -> str:
+    return format_statement(parse_sql(query)[0], parameters)
```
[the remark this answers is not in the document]
stepansergeevitch: This function was a temporary fix and is removed now

src/firebolt/async_db/cursor.py
```diff
@@ -164,24 +173,59 @@ def close(self) -> None:
...
-            self._descriptions = [
-                Column(d["name"], parse_type(d["type"]), None, None, None, None, None)
-                for d in query_data["meta"]
-            ]
-
-            # Parse data during fetch
-            self._rows = query_data["data"]
-        except (KeyError, JSONDecodeError) as err:
-            raise DataError(f"Invalid query data format: {str(err)}")
+        if response.headers.get("content-length", "") != "0":
```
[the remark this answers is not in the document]
stepansergeevitch: Yes. If no response is returned, `content-length` is explicitly set to `0` by server.
If there is no `content-length`, I think we should still try and parse the response.
Although, I think this situation shouldn't happen

src/firebolt/async_db/cursor.py
stepansergeevitch: Yes, seems better

src/firebolt/async_db/cursor.py
```diff
@@ -164,24 +173,59 @@ def close(self) -> None:
...
+            try:
+                query_data = response.json()
+                rowcount = int(query_data["rows"])
```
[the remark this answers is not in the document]
stepansergeevitch: Aggree

src/firebolt/async_db/_types.py
```diff
@@ -263,3 +258,23 @@ def process_token(token: Token) -> Token:
...
+    if not statements:
+        return [query]
+
+    if parameters:
+        if len(statements) > 1:
+            raise NotSupportedError(
+                "formatting multistatement queries is not supported"
+            )
+        return [format_statement(statements[0], paramset) for paramset in parameters]
+    return [str(st).strip().rstrip(";") for st in statements]
```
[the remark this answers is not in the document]
stepansergeevitch: Using strip is copied from sqparse `.split` implementation.
`rstrip` is used just for consistency, since some queries are parsed with ending ";" and some are without.

src/firebolt/async_db/cursor.py
ptiurin: Makes sense

src/firebolt/async_db/cursor.py
ptiurin: Got it, missed reset in _do_execute

yuryfirebolt (approved):

ptiurin (approved):

PR #111, opened 2022-01-20
fix: Added decimal.Decimal to _types.py.

ericf-firebolt:
As stated, added `decimal.Decimal` to `_types.py` to support dbt.


src/firebolt/async_db/_types.py
```diff
@@ -177,7 +178,7 @@ def parse_value(
     """Provided raw value and python type, parses first into python value."""
     if value is None:
         return None
-    if ctype in (int, str, float):
+    if ctype in (int, str, float, Decimal):
```
stepansergeevitch, line 181: I think that this condition is redundant.
We won't ever have ctype `Decimal`

stepansergeevitch (commented):
Please also add unit tests

src/firebolt/async_db/_types.py
ericf-firebolt: Done.

ericf-firebolt:
> Please also add unit tests

Decimal is a little weird/annoying in its inconsistency when converting numerical types vs strings, so I added two checks, one sending in a str, which returns a rather long string, and one sending in a numerical type—in this case a float—which returns a string with the same "significant digits" as the number that was input.

stepansergeevitch (approved):

PR #113, opened 2022-01-24
ci: Add mypy pre-commit hook

stepansergeevitch:
Added mypy check as a part of pre-commit hooks. Therefore, removed mypy step from code-check workflow


ptiurin (approved):
LGTM

yuryfirebolt (approved):

PR #115, opened 2022-01-24
feat: Implement nowait parameter for stop engine

yuryfirebolt:


src/firebolt/model/engine.py
```diff
@@ -38,6 +38,14 @@ class EngineKey(FireboltBaseModel):
     engine_id: str
 
 
+def wait(seconds: int, timeout_time: float, error_message: str, verbose: bool) -> None:
+    time.sleep(seconds)
+    if time.time() > timeout_time:
```
stepansergeevitch, line 43: I think it should be >=

stepansergeevitch (approved):

src/firebolt/model/engine.py
yuryfirebolt: Switching from `>` to `>=` would make a nanosecond difference

PR #116, opened 2022-01-24
fix: Setting TCP keepalive values

ptiurin:
https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout
ALB timeout is set to 350 seconds. If any connection is kept for longer than that it will be abandoned. This leads to long-running queries being stuck.
Those two settings set keepalive mechanism on TCP to avoid the above.
Fix can be tested using 
```
cursor.execute(
    "SELECT sleepEachRow(1) from numbers(360)", set_parameters={"advanced_mode": "1", "use_standard_sql": "0"}
)
```

https://github.com/firebolt-db/firebolt-python-sdk/pull/114 is a pre-requisite since httpcore has changed how it interacts with sockets and different way of overriding is needed with the old version.


ptiurin:
Unit tests will succeed after https://github.com/firebolt-db/firebolt-python-sdk/pull/114 is merged.

src/firebolt/async_db/connection.py
```diff
@@ -131,6 +134,28 @@ async def connect_inner(
...
+    """
+
+    async def connect_tcp(
+        self, host: str, port: int, timeout: float = None, local_address: str = None
+    ) -> AsyncNetworkStream:
+        stream = await super().connect_tcp(
+            host, port, timeout=timeout, local_address=local_address
+        )
+        # Enable keepalive
+        stream.get_extra_info("socket").setsockopt(SOL_SOCKET, SO_KEEPALIVE, 1)
```
stepansergeevitch, line 153: nit: Make constants into global variables

stepansergeevitch (commented):
Can we have an integration test for this case?

ptiurin:
> Can we have an integration test for this case?

We can, but it would significantly slow down integration tests execution. The test needs to run for 360 seconds.

src/firebolt/async_db/connection.py
```diff
@@ -17,6 +20,8 @@
 from firebolt.common.util import fix_url_schema
 
 DEFAULT_TIMEOUT_SECONDS: int = 5
+KEEPALIVE_FLAG: int = 1
+KEEPIDLE_RATE: int = 60  # seconds
```
yuryfirebolt, line 24: do I understand correctly, as long as this value is less than 350, the trick will work, right?

yuryfirebolt (commented):
Generally looks good

yuryfirebolt (approved):

src/firebolt/async_db/connection.py
```diff
@@ -131,6 +136,32 @@ async def connect_inner(
...
+    """
+    This class is a short-term solution for TCP keep-alive issue:
+    https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#connection-idle-timeout
+    Since httpx creates a connection right before executing a request
+    backend has to be overridden in order to set the socket KEEPALIVE
+    and KEEPIDLE settings.
+    """
+
+    async def connect_tcp(
+        self, host: str, port: int, timeout: float = None, local_address: str = None
```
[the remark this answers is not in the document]
ptiurin: httpcore didn't have them, but it's fair to add more correct typing. Added.

src/firebolt/async_db/connection.py
ptiurin: Technically yes, I was trying to find a typical expected value for this but it differs a lot. Wiki gives 45-60 range so I went with the latter.

ptiurin:
As discussed offline, added integration tests here anyway. Will look into parallelising the integration tests as a separate PR.

PR #117, opened 2022-01-25
ci: Parallelising the integration tests

ptiurin:
Making sure all tables created in integration tests are unique so they can be run in parallel without interfering with one another.


yuryfirebolt (approved):
Looks good

stepansergeevitch (approved):

PR #119, opened 2022-01-26
fix: MacOS keepalive setting

ptiurin:
TCP_KEEPIDLE is not available on MacOS. TCP_KEEPALIVE is the alternative, however it's not added to the `socket` library until Python 3.10. This is a workaround.

Tested locally on a mac.


ptiurin:
We have integration tests for this. Not sure if unit tests are appropriate, we can only check if correct functions are called on a mocked socket?

stepansergeevitch (approved):

PR #120, opened 2022-01-27
fix: Jupyter runtime

stepansergeevitch:
Fixed jupyter notebook async issues, extended dbapi jupyter examples.
This fix also add's an ability to use sync connection and cursor inside an asyncio loop, which we didn't support previously.
The issue was that basically loops can't be nested in asyncio, and all jupyter code is run inside a loop. The workaround for this is to create a thread inside a running loop and use new loop there. One problem is that a single `httpx.AsyncClient` should use the same loop always, so connection needs to cache it, which is done by storing `AsyncJobThread`

### DIFF
--- a/examples/dbapi.ipynb
+++ b/examples/dbapi.ipynb
@@ -37,15 +37,14 @@
    "source": [
     "# Only one of these two parameters should be specified\n",
     "engine_url = \"\"\n",
-    "engine_name = \"sburlakov_test_integration\"\n",
     "assert bool(engine_url) != bool(\n",
     "    engine_name\n",
     "), \"Specify only one of engine_name and engine_url\"\n",
     "\n",
-    "database_name = \"sburlakov_test\"\n",
-    "username = \"stepan.burlakov@firebolt.io\"\n",
-    "password = \"Not55566\"\n",
-    "api_endpoint = \"api.dev.firebolt.io\"  # DEFAULT_API_URL"
+    "database_name = \"\"\n",
+    "username = \"\"\n",
+    "password = \"\"\n",
+    "api_endpoint = DEFAULT_API_URL"
    ]
   },
   {
--- a/examples/dbapi.ipynb
+++ b/examples/dbapi.ipynb
@@ -10,7 +10,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "bdd3e404",
    "metadata": {},
    "outputs": [],
@@ -30,7 +30,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "id": "0ce8b2d3",
    "metadata": {},
    "outputs": [],
@@ -57,7 +57,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "id": "646869f7",
    "metadata": {},
    "outputs": [],
--- a/examples/dbapi.ipynb
+++ b/examples/dbapi.ipynb
@@ -10,13 +10,14 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "id": "bdd3e404",
    "metadata": {},
    "outputs": [],
    "source": [
     "from firebolt.db import connect\n",
-    "from firebolt.client import DEFAULT_API_URL"
+    "from firebolt.client import DEFAULT_API_URL\n",
+    "from datetime import datetime"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


examples/dbapi.ipynb
```diff
@@ -29,22 +30,22 @@
...
     "), \"Specify only one of engine_name and engine_url\"\n",
     "\n",
-    "database_name = \"\"\n",
-    "username = \"\"\n",
-    "password = \"\"\n",
-    "api_endpoint = DEFAULT_API_URL"
+    "database_name = \"sburlakov_test\"\n",
+    "username = \"stepan.burlakov@firebolt.io\"\n",
+    "password = \"Not55566\"\n",
+    "api_endpoint = \"api.dev.firebolt.io\"  # DEFAULT_API_URL"
```
ptiurin, line 48: Do you want to expose your email?

src/firebolt/db/connection.py
```diff
@@ -42,18 +42,18 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         # Holding this lock for write means that connection is closing itself.
         # cursor() should hold this lock for read to read/write state
         self._closing_lock = RWLockWrite()
+        self._async_job_thread = AsyncJobThread()
 
-    @wraps(AsyncBaseConnection.cursor)
     def cursor(self) -> Cursor:
         with self._closing_lock.gen_rlock():
-            c = super().cursor()
+            c = super()._cursor(async_job_thread=self._async_job_thread)
```
ptiurin, line 49: Would this work with multiple cursors from a single connection? And if multiple cursors are not supported can we lock it down somehow?

src/firebolt/common/util.py
```diff
@@ -37,15 +51,52 @@ def fix_url_schema(url: str) -> str:
     return url if url.startswith("http") else f"https://{url}"
 
 
-def async_to_sync(f: Callable) -> Callable:
+class AsyncJobThread:
+    def __init__(self) -> None:
+        self.loop: Optional[AbstractEventLoop] = None
+        self.result: Optional[Any] = None
+        self.exception: Optional[BaseException] = None
+        super().__init__()
```
ptiurin, line 59: What's `super()` in this context, since we're not parenting off anything? Object?

src/firebolt/common/util.py
```diff
@@ -37,15 +51,52 @@ def fix_url_schema(url: str) -> str:
...
+
+    def _initialize_loop(self) -> None:
+        if not self.loop:
+            try:
+                self.loop = get_event_loop()
+            except RuntimeError:
+                self.loop = new_event_loop()
```
ptiurin, line 66: I think part of the functionality of `get_event_loop` is that it will create a loop if one is not running. Would there be a case when `except:` here happens?

src/firebolt/common/util.py
```diff
@@ -37,15 +51,52 @@ def fix_url_schema(url: str) -> str:
     return url if url.startswith("http") else f"https://{url}"
 
 
-def async_to_sync(f: Callable) -> Callable:
+class AsyncJobThread:
+    def __init__(self) -> None:
```
ptiurin, line 55: Please add a docstring as to why we have this class in the first place. I have a feeling I'll forget in a few month time what was this about and a docstring would be helpful.

src/firebolt/db/connection.py
stepansergeevitch: Yes, the point is that every cursor need to share the same async_job_thread

examples/dbapi.ipynb
stepansergeevitch: Oh, I missed that, thanks

src/firebolt/common/util.py
stepansergeevitch: Yeah, that's redundant

src/firebolt/common/util.py
stepansergeevitch: This happens almost every time. Despite the description, it fails with runtime if no event loop is set

src/firebolt/common/util.py
stepansergeevitch: Yes, sure

src/firebolt/db/connection.py
ptiurin: Ok, got it.

src/firebolt/common/util.py
ptiurin: Interesting. Let's pop a quick comment there saying this happens despite get_event_loop claiming to do otherwise.

ptiurin (approved):
Apart from that LGTM

src/firebolt/common/util.py
```diff
@@ -37,15 +51,56 @@ def fix_url_schema(url: str) -> str:
     return url if url.startswith("http") else f"https://{url}"
 
 
-def async_to_sync(f: Callable) -> Callable:
+class AsyncJobThread:
+    """
+    Thread runner that allows running async tasks syncronously in a separate thread.
+    Caches loop to be reused in all threads
```
yuryfirebolt, line 57: Just one question, why do we need a class, that runs a function in another thread, but synchronously? I would add in comments, which exact problem this class is solving

yuryfirebolt (approved):
Looks good for me

PR #125, opened 2022-01-28
feat(FIR-10741): Defaulting to cheapest engine type

ptiurin:
In order to avoid relying on hardcoded defaults of the engine type we want to dynamically select a default instance if a user does not specify one. A cheapest instance in terms of cost is a good idea for testing or when this field is accidentally left unset.


src/firebolt/service/instance_type.py
```diff
@@ -41,6 +43,27 @@ def instance_types_by_name(self) -> Dict[InstanceTypeLookup, InstanceType]:
...
+        ]
+        # Filter out instances without storage
+        instance_list = [
+            i
+            for i in instance_types
+            if i.storage_size_bytes and i.storage_size_bytes != "0"
+        ]
+        cheapest = min(
+            instance_list,
+            key=lambda x: x.price_per_hour_cents if x.price_per_hour_cents else maxsize,
```
yuryfirebolt, line 63: I would better use `float('Inf')` instead of `maxsize`

src/firebolt/service/instance_type.py
```diff
@@ -41,6 +43,27 @@ def instance_types_by_name(self) -> Dict[InstanceTypeLookup, InstanceType]:
...
+        instance_types = [
+            InstanceType.parse_obj(i["node"]) for i in response.json()["edges"]
+        ]
+        # Filter out instances without storage
+        instance_list = [
+            i
+            for i in instance_types
+            if i.storage_size_bytes and i.storage_size_bytes != "0"
+        ]
+        cheapest = min(
```
yuryfirebolt, line 61: What if `instance_list` is empty. We should raise something more informative than `ValueError: min() arg is an empty sequence`

yuryfirebolt (approved):
Looks good, just a few minor comments

stepansergeevitch (approved):

src/firebolt/service/instance_type.py
ptiurin: Yeah, good suggestion

src/firebolt/service/instance_type.py
ptiurin: Fair point, added a better error in here. Looks like this whole code could use better error messaging. Hopefully the API rework comes soon enough.

PR #128, opened 2022-02-01
feat: Fir 10511 feature request support resour

stepansergeevitch:
Allowing users to use access token for authentication in both `ResourceManager` and dbapi `Connection`.
- Added `Auth.from_token()`
- Added `Settings.access_token`
- Added `access_token` parameter to dbapi `connect` function

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -8,9 +8,10 @@ from typing import Any, Callable, List, Optional, Type
 from httpcore.backends.auto import AutoBackend
 from httpcore.backends.base import AsyncNetworkStream
 from httpx import AsyncHTTPTransport, HTTPStatusError, RequestError, Timeout
+from httpx._types import AuthTypes
 
 from firebolt.async_db.cursor import BaseCursor, Cursor
-from firebolt.client import DEFAULT_API_URL, AsyncClient
+from firebolt.client import DEFAULT_API_URL, AsyncClient, Auth
 from firebolt.common.exception import (
     ConnectionClosedError,
     FireboltEngineError,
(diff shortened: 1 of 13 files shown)


src/firebolt/async_db/connection.py
```diff
@@ -64,11 +64,43 @@ async def _resolve_engine_url(
...
+def _get_auth(
+    username: Optional[str], password: Optional[str], access_token: Optional[str]
+) -> AuthTypes:
+    if not access_token:
+        if not username or not password:
+            raise InterfaceError(
+                "Neither username/password nor access_token are provided. Provide one"
+                " to authenticate"
+            )
+        return (username, password)
```
ptiurin, line 89: Should we create an Auth instance here with user/password instead of passing a tuple?

tests/unit/async_db/test_connection.py
```diff
@@ -90,18 +90,60 @@ async def test_cursor_initialized(
...
+        assert await cursor.execute("select*") == -1
+
+    with raises(InterfaceError) as exc_info:
+        async with await connect(engine_url="engine_url", database="database"):
+            pass
+    assert (
+        str(exc_info.value)
+        == "Neither username/password nor access_token are provided. Provide one to"
+        " authenticate"
+    )
```
ptiurin, line 131: I'm not a fan of checking the exception strings. As long as the correct exception class is raised we should be ok. This way if we were to fix typos or reword the exceptions we wouldn't have to change tests.
In this case I'd actually create a new error class, something like SettingsError, since from the description InterfaceError is not exactly what we want here.

ptiurin (approved):
Generally lgtm, just a couple of suggestions.

src/firebolt/async_db/connection.py
stepansergeevitch: To create an Auth instance here we would need to also know `api_endpoint` here. In case of tuple, it's inherited from Client

ptiurin: I don't have a strong opinion in here, just thought it would look neater.

PR #129, opened 2022-02-01
feat: Implement database update method

yuryfirebolt:


src/firebolt/model/database.py
```diff
@@ -121,3 +125,40 @@ def delete(self) -> Database:
...
+            database: Database
+            database_id: str
+            update_mask: FieldMask
+
+        logger.info(
+            f"Updating Database (database_id={self.database_id}, "
+            f"name={self.name}, description={self.description})"
+        )
+
+        payload = _DatabaseUpdateRequest(
```
stepansergeevitch, line 147: Do you think we can get rid of `_DatabaseUpdateRequest` and just create a plain dict here?
Do we have any profit of using it?

stepansergeevitch: It just seems that it's redundant here

stepansergeevitch (commented):
Overall lgtm, left single question

stepansergeevitch (approved):

PR #131, opened 2022-02-02
feat: Implement engine restart method

yuryfirebolt:


tests/unit/service/conftest.py
```diff
@@ -214,7 +214,7 @@ def do_mock(
         request: httpx.Request = None,
         **kwargs,
     ) -> Response:
-        assert request.url == engine_url
+        # assert request.url == engine_url
```
stepansergeevitch, line 217: I assume you removed this check because of `:restart` suffix? If so, please remove the suffix before check. Or just remove the comment if you think it's redundant

stepansergeevitch (approved):

ptiurin (approved):

tests/unit/service/conftest.py
yuryfirebolt: Yes, will do

PR #132, opened 2022-02-02
fix: backwards compatibility for resource manager

ptiurin:
`account_name` was an unused parameter that was removed in 0.4.0 . However, some sdk users passed it in and a new release breaks that. Adding the parameter back and providing a warning that it should not be used.

### DIFF
--- a/src/firebolt/service/manager.py
+++ b/src/firebolt/service/manager.py
@@ -1,4 +1,5 @@
 from typing import Optional
+from warnings import warn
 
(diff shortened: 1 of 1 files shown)


ptiurin:
As discussed during the daily, we're below 1.0.0 so breaking changes could happen. We'll see if anyone else has issue with the same thing and if too many users are affected roll out a backwards-compatible fix. For now it highlights the use of incorrect setting that does nothing.

PR #133, opened 2022-02-03
feat: Implement engine update

yuryfirebolt:


src/firebolt/model/engine.py
```diff
@@ -320,6 +325,95 @@ def stop(
...
+                engine_id=self.engine_id, account_id=self._service.account_id
+            ),
+            headers={"Content-type": "application/json"},
+            json=_EngineUpdateRequest(
+                account_id=self._service.account_id,
+                engine=self,
+                desired_revision=desired_revision,
+                engine_id=self.engine_id,
+                update_mask=FieldMask(
+                    paths=list(
```
stepansergeevitch, line 398: nit: Can we calculate paths in a separate function, or at least take it out from patch request?

src/firebolt/model/engine.py
```diff
@@ -320,6 +325,95 @@ def stop(
...
+        auto_stop: Optional[int] = None,
+        warmup: Optional[WarmupMethod] = None,
+        description: Optional[str] = None,
+    ) -> Engine:
+        """
+        update the engine, and returns an updated version of the engine, all parameters
+        could be set to None, this would keep the old engine parameter value
+        """
+
+        class _EngineUpdateRequest(FireboltBaseModel):
```
stepansergeevitch, line 343: Same question as for a database update PR: can we have a plain dict here?

yuryfirebolt: Generally it is possible, but the code will get a bit uglier, since, we would need to call `.jsonable_dict()` for the each dict field, which is `Object` and additionally check whether it is not `None`. Currently, it is all done automatically within `FireboltBaseModel`

PR #135, opened 2022-02-09
feat(FIR-10892): Internal args for engine

ptiurin:
Allowing special parameters like `db_version`, `proxy_version` which are part of the API, but should be obfuscated from the external user. As discussed we're fine if users find out about them via reading the code, but reading the docs should not give hints about their existence.
This will mainly be used for automated testing by internal teams.

### DIFF
--- a/src/firebolt/service/engine.py
+++ b/src/firebolt/service/engine.py
@@ -1,5 +1,5 @@
 from logging import getLogger
-from typing import List, Optional, Union
+from typing import Any, Dict, List, Optional, Union
 
 from firebolt.common.exception import FireboltError
 from firebolt.common.urls import (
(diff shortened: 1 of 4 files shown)


src/firebolt/service/engine.py
```diff
@@ -186,11 +188,8 @@ def create(
             specification=EngineRevisionSpecification(
                 db_compute_instances_type_key=instance_type_key,
                 db_compute_instances_count=scale,
-                db_compute_instances_use_spot=False,
```
stepansergeevitch, line 189: Do we wan't to have some defaults for `db_compute_instances_use_spot`, `db_version`, `proxy_instances_count` and `proxy_version`?

stepansergeevitch (approved):

src/firebolt/service/engine.py
ptiurin: We already have those defaults in `EngineRevisionSpecification`, https://github.com/firebolt-db/firebolt-python-sdk/blob/edf5df76ad59ce0192c3c34ff9816542d0621884/src/firebolt/model/engine_revision.py#L34

PR #145, opened 2022-02-22
ci: Relative workflow paths

ptiurin:
GitHub recently fixed relative workflow reference. It pulls whichever version of the dependent workflow is present on the current branch. Meaning we no longer need to swap workflow path each time we want to test a change on it on a branch 
e.g. before - any changes to security-scan.yml would have to be tested by changing pull-request.yml to call security-scan.yml@<branchname> and then back to @ main. Now none of this is needed as workflows reference changes on a branch.


yuryfirebolt (approved):
great, it would make stuff a bit easier

PR #146, opened 2022-02-23
perf: Implement mechanism for caching token on login

yuryfirebolt:
On the first successful authentication, get the token and save it in app_config_dir token file. The token itself is encrypted using the username and password tuple. Hence, the token cannot be accessed, by someone who doesn’t know the username and the password.

During establishing the connection, we check, whether something is written in the app_config_dir token file. If yes, we decrypt it using the username and password. In case of successful decryption, we use this token for connection. If the connection failed, we use username, password as a fallback option.

For the encryption/decryption symmetric encryption Fernet is used with cryptography python library.


src/firebolt/common/token_storage.py
```diff
@@ -0,0 +1,114 @@
...
+        """
+        Class for permanent storage of token in the filesystem in encrypted way
+
+        :param username: username used for toke encryption
+        :param password: password used for toke encryption
+        """
+        self._data_dir = user_data_dir(appname="firebolt")
+        os.makedirs(self._data_dir, exist_ok=True)
+
+        self._token_file = os.path.join(self._data_dir, "token.json")
```
stepansergeevitch, line 24: Seems like this implementation only supports caching token for a single set of credentials.
I would suggest changing `token.json` to something like `hash(username + password).json`, which would resolve this issue

src/firebolt/common/token_storage.py
```diff
@@ -0,0 +1,114 @@
...
+
+        :return: token or None
+        """
+        res = self._read_data_json()
+        if "token" not in res:
+            return None
+
+        return self.encrypter.decrypt(res["token"])
+
+    def cache_token(self, token: str) -> None:
```
stepansergeevitch, line 70: It could be useful to also store token expiration here. This way we can reduce the amount of requests by skipping expired tokens

src/firebolt/common/token_storage.py
```diff
@@ -0,0 +1,114 @@
...
+            iterations=39000,
+        )
+        self.fernet = Fernet(
+            base64.urlsafe_b64encode(
+                kdf.derive(bytes(f"{username}{password}", encoding="utf-8"))
+            )
+        )
+
+    @staticmethod
+    def generate_salt() -> str:
```
stepansergeevitch, line 104: nit: separate method

tests/unit/common/test_token_storage.py
```diff
@@ -0,0 +1,104 @@
...
+    TokenSecureStorage(**settings).cache_token(token)
+
+    assert token == TokenSecureStorage(**settings).get_cached_token()
+    token = "some new string to encrypt"
+
+    TokenSecureStorage(**settings).cache_token(token)
+    assert (
+        token
+        == TokenSecureStorage(
+            username="username", password="password"
```
ptiurin, line 66: nit: `**settings` like in others

stepansergeevitch (commented):
Overall looks good

src/firebolt/client/resource_manager_hooks.py
```diff
@@ -40,14 +40,24 @@ def raise_on_4xx_5xx(response: Response) -> None:
             parsed_response = exc.response.json()
         except JSONDecodeError:
             parsed_response = {"_raw": exc.response.text}
+
+        if exc.response.status_code == codes.UNAUTHORIZED:
+            # Do not raise an exception on UNAUTHORIZED
+            # It should be responsibility of authentication mechanism
+            return
```
ptiurin, line 47: What if the token is expired after auth and before command execution?

.github/workflows/code-check.yml
```diff
@@ -26,3 +26,7 @@ jobs:
       uses: pre-commit/action@v2.0.3
       with:
         extra_args: --all-files
+
+    - name: Type check with mypy
```
stepansergeevitch, line 30: Mypy checks are already performed inside pre-commit

ptiurin (approved):
LGTM

tests/conftest.py
```diff
@@ -0,0 +1,8 @@
+from pyfakefs.fake_filesystem_unittest import Patcher
+from pytest import fixture
+
+
+@fixture(autouse=True)
+def global_fake_fs() -> None:
+    with Patcher():
+        yield
```
ptiurin, line 8: To be honest, I preferred explicitly adding fs to the relevant tests. That way there's no hidden functionality, you know which tests are hitting a fake file system and which do not.
But up to you, I'm not too worried about it.

stepansergeevitch: This is required because now any authentication leaves token cache artifact file, which also causes tests to fail

stepansergeevitch: We would need to add fake fs to almost every integration test and to a lot of unit tests

PR #147, opened 2022-03-01
fix expected nullable type

stepansergeevitch:
Fixed failing integration tests
For some reason, we were selecting `nullable(int)` and receiving `nullable(str)`. This seems to be fixed now
Test run:
https://github.com/firebolt-db/firebolt-python-sdk/runs/5377713096?check_suite_focus=true


ptiurin (approved):
lgtm

PR #148, opened 2022-03-09
feat:Fir 11238 python sdk support new data types

stepansergeevitch:
Added support for `decimal`, `date32` and `datetime64` types.
Added unit tests and for decimal added integration tests


src/firebolt/async_db/__init__.py
```diff
@@ -2,6 +2,8 @@
     ARRAY,
     BINARY,
     DATETIME,
+    DATETIME64,
+    DECIMAL,
     NUMBER,
```
ptiurin, line 7: Do we want DATE32 here as well?

src/firebolt/async_db/_types.py
```diff
@@ -145,20 +181,38 @@ def python_type(self) -> type:
...
     # Handle arrays
     if raw_type.startswith(ARRAY._prefix) and raw_type.endswith(")"):
         return ARRAY(parse_type(raw_type[len(ARRAY._prefix) : -1]))
+    # Handle decimal
+    if raw_type.startswith(DECIMAL._prefix) and raw_type.endswith(")"):
+        try:
+            prec_scale = raw_type[len(DECIMAL._prefix) : -1].split(",")
+            precision, scale = int(prec_scale[0]), int(prec_scale[1])
+        except (ValueError, IndexError):
+            pass
```
ptiurin, line 205: What happens in this case?  Should we raise those instead?

src/firebolt/async_db/_types.py
```diff
@@ -93,6 +93,41 @@ def __eq__(self, other: object) -> bool:
...
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, DECIMAL):
+            return NotImplemented
+        return other.precision == self.precision and other.scale == self.scale
+
+
+class DATETIME64:
+    """Class for holding imformation about datetime64 value in firebolt db."""
+
+    _prefix = "DateTime64("
```
ptiurin, line 117: Is `_prefix` used anywhere?

tests/unit/async_db/conftest.py
```diff
@@ -61,12 +75,15 @@ def query_data() -> List[List[ColType]]:
             922337203685477580,
             -922337203685477580,
             1,
-            1.0387398573,
+            "1.0387398573",
             "some text",
             "2019-07-31",
+            "2020-07-31",
```
ptiurin, line 81: New dates extended the range below 1970 so let's reflect this in a test (despite it being a unit test) e.g. `1860-01-31`, which was not possible with regular DATE

tests/unit/async_db/test_typing_parse.py
```diff
@@ -92,8 +94,8 @@ def test_parse_value_datetime() -> None:
         assert str(exc_info.value) == f"Invalid date value {value}: str expected"
 
     # Datetime
-    assert parse_value("2021-12-31 23:59:59", datetime) == datetime(
-        2021, 12, 31, 23, 59, 59
```
ptiurin, line 96: Should the same test be there for date32?

ptiurin (commented):
We should probably add integration tests for extended date and timestamps?

src/firebolt/async_db/_types.py
stepansergeevitch: In case we're not able to parse the type, we always return `str`. That is what's going to happen here

tests/unit/async_db/test_typing_parse.py
stepansergeevitch: date32 doesn't support milliseconds

tests/unit/async_db/conftest.py
stepansergeevitch: Sure, agree

tests/unit/async_db/test_typing_parse.py
stepansergeevitch: Oh, I see what you mean

src/firebolt/async_db/_types.py
ptiurin: Found it, yeah.

src/firebolt/async_db/_types.py
ptiurin: I see, I guess the only issue would be a useless call to _InternalType, since we already established it's a decimal, but I can live with that.

ptiurin (approved):
lgtm

PR #149, opened 2022-03-16
feat: Fir 11986 full support for set statement

stepansergeevitch:
Added support for set statements based on the design doc
https://docs.google.com/document/u/1/d/1tuo_zAbRYUog3OiW8duMYkivsQhi20gutCibxa5o5ic/edit
Added unit and integration tests.
Added deprecation warning for old way to pass set parameters


src/firebolt/async_db/cursor.py
```diff
@@ -260,39 +235,105 @@ def _reset(self) -> None:
...
+                ] = (-1, None, None)
+                if isinstance(query, Set):
+                    # Validate parameter by executing simple query with it
+                    resp = await self._api_request(
+                        "select 1", {query.name: query.value}
+                    )
+                    # Handle invalid set parameter
+                    if resp.status_code == codes.BAD_REQUEST:
+                        raise OperationalError(resp.text)
+                    await self._raise_if_error(resp)
```
ptiurin, line 326: Wouldn't `_raise_if_error` be sufficient here? Why do we need L324-325?

src/firebolt/async_db/cursor.py
```diff
@@ -260,39 +235,105 @@ def _reset(self) -> None:
...
+                "Passing set parameters as an argument is deprecated. Please run "
+                "a query 'SET <param> = <value>'"
+            )
         try:
 
-            queries = split_format_sql(query, parameters)
+            queries = split_format_sql(raw_query, parameters)
 
-            for query in queries:
+            for query in queries:  # type: ignore
```
ptiurin, line 303: Just curious, what was causing mypy issues here?

tests/unit/async_db/test_cursor.py
```diff
@@ -475,3 +476,82 @@ async def test_cursor_multi_statement(
...
+    with raises(DataError):
+        await cursor.fetchall()
+
+    assert (
+        len(cursor._set_parameters) == 1
+        and "param" in cursor._set_parameters
+        and cursor._set_parameters["param"] == "1"
+    )
+
+    cursor.flush_parameters()
```
ptiurin, line 529: Can we also add a test for multiple set paramter queries? Either executed separately or in the same statement. So then in the end `_set_parameters` contains both `a` and `b`.

ptiurin (approved):
Apart from those points, logic LGTM.

src/firebolt/async_db/_types.py
```diff
@@ -323,9 +332,41 @@ def process_token(token: Token) -> Token:
     return formatted_sql
 
 
+Set = namedtuple("Set", ["name", "value"])
```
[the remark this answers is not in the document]
stepansergeevitch: I'm thinking of `ParameterSet`

src/firebolt/async_db/cursor.py
stepansergeevitch: To return `OperationalError` with message in case of an invalid parameter. Inside of `_raise_if_error`, we call `raise_for_status`, which only returns error code, and is sufficient in most cases except for this one

src/firebolt/async_db/cursor.py
stepansergeevitch: This is redundant

tests/unit/async_db/test_cursor.py
stepansergeevitch: Yes, makes sense

src/firebolt/async_db/_types.py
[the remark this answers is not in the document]
ptiurin: How about `SetParameter` then? This way it sounds just how we call them outside of code.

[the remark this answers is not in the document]
stepansergeevitch: Sure, fine with me

PR #150, opened 2022-03-23
fix: Compatibility with cryptography 2.*

stepansergeevitch:
According to cryptograpy 3.1 changelog
```
**backend** arguments to functions are no longer required and the default backend will automatically be selected if no backend is provided.
```
In order to be compatible with cryptography 2.8, required by Redash, we should provide default backend as an argument

### DIFF
--- a/src/firebolt/common/token_storage.py
+++ b/src/firebolt/common/token_storage.py
@@ -9,6 +9,9 @@ from typing import Optional
 
 from appdirs import user_data_dir
 from cryptography.fernet import Fernet, InvalidToken
+from cryptography.hazmat.backends.openssl.backend import (
+    backend as ossl_backend,
+)
 from cryptography.hazmat.primitives import hashes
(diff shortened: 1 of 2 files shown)


src/firebolt/common/token_storage.py
```diff
@@ -114,6 +117,7 @@ def __init__(self, salt: str, username: str, password: str):
             salt=b64decode(salt),
             length=32,
             iterations=39000,
+            backend=ossl_backend,
```
ptiurin, line 120: Should we be using `default_backend()` function instead? At least that's what cryptography looks to suggest https://cryptography.io/en/3.1/hazmat/backends/#cryptography.hazmat.backends.default_backend

stepansergeevitch: Yeah, I think we can do this. I took an open ssl backen from current version implementation
https://github.com/pyca/cryptography/blob/36.0.x/src/cryptography/hazmat/primitives/kdf/pbkdf2.py#L52

ptiurin (approved):

PR #151, opened 2022-03-23
build: Downgrade the requirement for cryptography

ptiurin:
We can relax the requirement since we don't use any of the new functionality later versions offer. This will enable users with older requirements to be compatible.


yuryfirebolt (approved):

ptiurin:
Setting on hold, FOSSA found a security issue after all.

ptiurin:
Changing to 3.4 as it has the 2.8 vulnerability fixed. This is the lowest we can go according to the security scan.

PR #152, opened 2022-03-30
feat: Fir 12126 make engine parameter optional

stepansergeevitch:
Allow users to skip engine parameter. Get default engine url in this case.

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -83,7 +83,6 @@ async def _get_database_default_engine_url(
     ) as client:
         try:
             account_id = await client.account_id
-            # Get database id by name
             response = await client.get(
                 url=ACCOUNT_ENGINE_URL_BY_DATABASE_NAME.format(account_id=account_id),
                 params={"database_name": database},
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -50,10 +50,9 @@ async def _resolve_engine_url(
             )
             response.raise_for_status()
             engine_id = response.json()["engine_id"]["engine_id"]
-
             response = await client.get(
                 url=ACCOUNT_ENGINE_URL.format(
-                    account_id=(await client.account_id), engine_id=engine_id
+                    account_id=account_id, engine_id=engine_id
                 ),
             )
             response.raise_for_status()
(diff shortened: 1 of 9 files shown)


ptiurin:
Lgtm, just wondering if we can use `https://api.app.firebolt.io/core/v1/account/engines:getURLByDatabaseName` ? Looks like it returns a default engine url, without the extra steps.

stepansergeevitch:
> Lgtm, just wondering if we can use `https://api.app.firebolt.io/core/v1/account/engines:getURLByDatabaseName` ? Looks like it returns a default engine url, without the extra steps.

Oh, that actually seems to be much better, didn't see that endpoint. Will use this one

ptiurin (approved):
lgtm

src/firebolt/async_db/connection.py
```diff
@@ -136,7 +166,15 @@ async def connect_inner(
         # Mypy checks, this should never happen
         assert database is not None
 
-        if engine_name:
+        if not engine_name and not engine_url:
```
yuryfirebolt, line 169: Change the documentation on L153 to reflect the changes

src/firebolt/async_db/connection.py
```diff
@@ -65,17 +69,43 @@ async def _resolve_engine_url(
...
+            response.raise_for_status()
+            return response.json()["engine_url"]
+        except (
+            JSONDecodeError,
+            RequestError,
+            RuntimeError,
+            HTTPStatusError,
+            KeyError,
+        ) as e:
+            raise InterfaceError(f"Unable to retrieve default engine endpoint: {e}.")
```
yuryfirebolt, line 99: What if the real error is wrong credentials or expired/incorrect token?

yuryfirebolt (changes requested):

src/firebolt/async_db/connection.py
stepansergeevitch: Sure, we could also catch `AuthenticationError`

stepansergeevitch: On the other side, I think `AuthenticationError` is pretty representative itself, so I think we should just let it propagate further

yuryfirebolt (approved):

PR #154, opened 2022-04-04
docs(FIR-12444): Add CI badges to readme

ptiurin:
Adding CI badges to indicate code quality standards to repo contributors. 

Coverage is extracted from the report, uploaded to a gist, which is then referenced in the Readme. This ensures there's no need to commit an update to the readme to see the coverage changing.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -45,7 +45,10 @@ jobs:
       run: |
         fraction=$(sed -n 2p coverage.xml | sed 's/.*line-rate=\"\([0-9.]*\)\".*$/\1/')
         percentage=$(echo "scale=1; $fraction * 100" | bc -l)
-        echo "::set-output name=covered::${percentage%.*}"
+        percentage_whole=$(echo "${percentage%.*}")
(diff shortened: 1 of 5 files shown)


stepansergeevitch (approved):
LGTM

.github/workflows/unit-tests.yml
```diff
@@ -36,3 +39,27 @@ jobs:
...
+        echo "::set-output name=colour::$colour"
+        echo "::set-output name=covered::$percentage_whole"
+    
+    - name: Create Coverage Badge
+      uses: schneegans/dynamic-badges-action@v1.2.0
+      if: github.event_name == 'push'
+      continue-on-error: true
+      with:
+        auth: ${{ secrets.GIST_PAT }}
+        gistID: 65d5a42849fd78f4c6e62fad18490d20
```
yuryfirebolt, line 61: would it be possible to put it to secrets as well?

ptiurin: We can, but this is not particularly sensitive information, it just uniquely identifies the Gist, like file name. Also, it's not reusable like PAT. Therefore I'd rather keep it as is, avoiding overcomplicating the secrets.

PR #158, opened 2022-04-13
docs: Fir 12403 add linter rule requiring docs

stepansergeevitch:
Added new flake8 checkers for docstrings:
- flake8-docstrings - validate that every required object has a docsting.
- flake8-spellcheck - spelling checker for docs
- darglint - validated that docstings have all required items (args, return value, raised exceptions)

Added docstrings for client and common modules. Other modules will be covered in next PRs

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,6 +73,7 @@ warn_untyped_fields = True
 [flake8]
 max-line-length = 88
 per-file-ignores = __init__.py:F401
-ignore = E203, W503
+ignore = E203, W503, D100
 ban-relative-imports = True
+docstring-convention = google
 inline-quotes = "
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ warn_untyped_fields = True
 [flake8]
 max-line-length = 88
 per-file-ignores = __init__.py:F401
-ignore = E203, W503, D100, D104, D107
+ignore = E203, W503, D100, D104, D105, D107
 ban-relative-imports = True
 docstring-convention = google
 inline-quotes = "
--- a/setup.cfg
+++ b/setup.cfg
@@ -73,7 +73,7 @@ warn_untyped_fields = True
 [flake8]
 max-line-length = 88
 per-file-ignores = __init__.py:F401
-ignore = E203, W503, D100, D104, D105, D107
+ignore = E203, W503, D100, D104, D105, D107, SC200
 ban-relative-imports = True
 docstring-convention = google
 inline-quotes = "
--- a/setup.cfg
+++ b/setup.cfg
@@ -74,7 +74,14 @@ warn_untyped_fields = True
 exclude = tests/*
 max-line-length = 88
 per-file-ignores = __init__.py:F401
-ignore = E203, W503, D100, D104, D105, D107, SC200
+ignore =
+    E203, # Whitespace before ':'. Ignored for black compatibility
+    W503, # Line break occurred before a binary operator. Ignored for black compatibility
+    D100, # Missing docstring in public module
+    D104, # Missing docstring in public package
+    D105,# Missing docstring in magic method
+    D107, # Missing docstring in __init__
+    SC200, # Spelling error in name (e.g. variable, function, class)
 ban-relative-imports = True
 docstring-convention = google
 inline-quotes = "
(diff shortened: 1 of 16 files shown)


setup.cfg
```diff
@@ -73,6 +73,7 @@ warn_untyped_fields = True
 [flake8]
 max-line-length = 88
 per-file-ignores = __init__.py:F401
-ignore = E203, W503
+ignore = E203, W503, D100, D104, D105, D107, SC200
```
ptiurin, line 77: I can't seem to find those error codes. Where are they listed?

src/firebolt/client/auth.py
```diff
@@ -12,16 +12,26 @@
...
+    Attributes:
+        username (str): Username
+        password (str): Password
     """
 
     __slots__ = (
         "username",
         "password",
-        "api_url",
+        "_api_endpoint",
```
ptiurin, line 34: Love that we spotted this! However, I thought it would fail on runtime if an instance variable was not declared in __slots__ ? How was this working before?

src/firebolt/client/auth.py
```diff
@@ -50,18 +68,40 @@ def __init__(
...
+            Optional[str]: Acquired token
+        """
         return self._token
 
     @property
     def expired(self) -> Optional[int]:
+        """Check if current token is expired.
+
+        Returns:
+            bool: True if expired, False otherwise
```
ptiurin, line 92: Looks like it does return boolean, but our type hint is wrong.

src/firebolt/client/client.py
```diff
@@ -30,34 +32,51 @@ def __init__(
...
-    httpx.Client:
-
-    + (HttpxClient.__doc__ or "")
     """
 
     @cached_property
     def account_id(self) -> str:
+        """User account id.
+
+        If account_name is provided, returns it's id.
```
ptiurin, line 71: I'd rephrase it to something like "If account_name was provided during the instantiation of this class...".
This way we avoid confusion that the docstring talks about a parameter that was removed from `account_id()` and the comment was not updated.

src/firebolt/client/resource_manager_hooks.py
```diff
@@ -7,32 +7,53 @@
...
+    """Log HTTP requests.
+
+    Hook for an HTTP client
+
+    Args:
+        request (Request): Request to log
+    """
     logger.debug(
-        f"Request event hook: {request.method} {request.url} - Waiting for response"
+        "Request event hook: %s %s - Waiting for response", request.method, request.url
```
ptiurin, line 18: Is this the best practice, as opposed to f-strings, according to the checker?

src/firebolt/common/token_storage.py
```diff
@@ -126,9 +154,27 @@ def __init__(self, salt: str, username: str, password: str):
...
         return self.fernet.encrypt(bytes(data, encoding="utf-8")).decode("utf-8")
 
     def decrypt(self, data: str) -> Optional[str]:
+        """Decrypt encrypted data.
+
+        Args:
+            data (str): Encrypted data
+
+        Returns:
+            str: Decrypted data
```
ptiurin, line 175: probably need `optional` here as well, for consistency

setup.cfg
stepansergeevitch: Ds are related to darglint, SCs are related to spellcheck

src/firebolt/client/auth.py
stepansergeevitch: This is actually a tricky one.
The reason seems to be that base Auth class doesn't have `__slots__` defined, and therefore it creates a `__dict__` entry, which makes `__slots__` redundant. I think in this case we can leave it just for documentation purposes

src/firebolt/client/auth.py
stepansergeevitch: Hm, wondering how mypy didn't catch this

src/firebolt/client/client.py
stepansergeevitch: Makes sense!

src/firebolt/client/resource_manager_hooks.py
stepansergeevitch: Yes, this way it performs a lazy evaluation

ptiurin (approved):
LGTM, thanks for adding descriptions

PR #161, opened 2022-04-27
fix: engine incorrect auto_stop update in case it is 0

yuryfirebolt:


src/firebolt/model/engine.py
```diff
@@ -354,7 +354,9 @@ class _EngineUpdateRequest(FireboltBaseModel):
         # Update engine settings
         engine_settings_params = {
             "engine_type": engine_type,
-            "auto_stop_delay_duration": f"{auto_stop * 60}s" if auto_stop else None,
+            "auto_stop_delay_duration": f"{auto_stop * 60}s"
+            if auto_stop is not None
```
ptiurin, line 358: While we're at it should we exclude negative numbers as well?

ptiurin (approved):
Apart from that suggestion, lgtm

src/firebolt/model/engine.py
yuryfirebolt: So far we don't do any input validation. It is just when auto_stop was set to zero, which is a valid value, the if statement didn't allow to propagate it.

PR #164, opened 2022-05-12
fix: Added quotes to "firebolt-sdk[ciso8601]"

marknoack:
Original pip commands had incorrect string, needed double-quotes


stepansergeevitch:
For me, this command works fine without quotes as well. Do you have any issues with it?

FireboltCorey:
Security check fail not a blocker.

FireboltCorey (approved):

PR #166, opened 2022-05-13
refactor: Fir 13300 allow providing generic auth

stepansergeevitch:
Preparations for adding service accounts authentication method:
- Create base Auth class, as well as separate classes for Username/Password and Token authentication
- Allow to provide Auth object instead of raw credentials. Deprecate the last.
- Remove `api_endpoint` from auth objects. It's now only stored in Client, which is responsible for updating authentication requests base url


src/firebolt/client/auth/username_password.py
```diff
@@ -0,0 +1,112 @@
...
+        Yields:
+            Request: An http request to get token. Expects Response to be sent back
+
+        Raises:
+            AuthenticationError: Error while authenticating with provided credentials
+        """
+        try:
+            response = yield self.request_class(
+                "POST",
+                AUTH_URL,
```
ptiurin, line 82: Maybe pop a comment here that the final URL is generated in the client.

tests/unit/client/auth/test_auth.py
```diff
@@ -0,0 +1,126 @@
...
+    httpx_mock: HTTPXMock,
+    test_username: str,
+    test_password: str,
+    test_token,
+) -> None:
+    # Hack to replace Auth._token_cache attribute
+    class MAuth(Auth):
+        @property
+        def _token_storage(self):
+            return TokenSecureStorage(test_username, test_password)
```
ptiurin, line 96: Would `unittest.mock.patch` be a better way of doing this?

ptiurin (approved):
Apart from minor comments, lgtm.

src/firebolt/client/auth/username_password.py
stepansergeevitch: Will do

tests/unit/client/auth/test_auth.py
stepansergeevitch: Yeah, didn't think of it for some reason

PR #172, opened 2022-05-31
feat(FIR-9997): Usage tracking

ptiurin:
Tracking usage based on parsing the stack and detecting known integration signatures of file + function name.
Also allowing access for manual override/addition of connector information.

Addition of a UsageTracker instance to the Conection causes a diffferent garbage collection to trigger, therefore a warnings test also had to be reworked with a manual call to gc. Users still get a warning of an unclosed connection if their script exits without doing so. This change only affects how we test the warnings, not how the users of the SDK see them.


src/firebolt/utils/usage_tracker.py
```diff
@@ -0,0 +1,159 @@
...
+from platform import python_version, release, system
+from sys import modules
+from typing import Dict, List, Optional, Tuple, Union
+
+from firebolt import __version__
+
+logger = logging.getLogger(__name__)
+
+
+class Format(Enum):
```
stepansergeevitch, line 13: Why do we need this type? I believe if we'll switch to another tracking method, we'll just rewrite the logic

src/firebolt/utils/usage_tracker.py
```diff
@@ -0,0 +1,159 @@
...
+        "Python %s detected. SDK %s OS %s %s",
+        py_version,
+        sdk_version,
+        os_version,
+        ciso,
+    )
+    return (py_version, sdk_version, os_version, ciso)
+
+
+class UsageTracker:
```
stepansergeevitch, line 82: Can we remove this class completely, transforming it into a function?

src/firebolt/async_db/connection.py
```diff
@@ -285,17 +286,19 @@ def __init__(
         database: str,
         auth: Auth,
         api_endpoint: str = DEFAULT_API_URL,
+        connector_versions: Union[Tuple[str, str], List[Tuple[str, str]]] = None,
```
stepansergeevitch, line 289: We usually don't create a `Connection` directly, but rather using the `connect` function. Can we add a parameter there as well?
Speaking of having too many parameters there, we could think of just adding an `additional_parameters: dict` argument, where we could later put anything, not used often

ptiurin: You're right, I missed this part.

src/firebolt/utils/usage_tracker.py
ptiurin: Ok, you're probably right. I wanted to encapsulate the functionality a bit better.

src/firebolt/utils/usage_tracker.py
ptiurin: I'm not sure if we'll get rid of the user-agent completely, but sure, let's leave that for the time when decision needs to be made.

src/firebolt/utils/usage_tracker.py
```diff
@@ -0,0 +1,144 @@
...
+
+def detect_connectors() -> Dict[str, str]:
+    """
+    Detect which connectors are running the code by parsing the stack.
+    Exceptions are ignored since this is intended for logging only.
+    """
+    connectors: Dict[str, str] = {}
+    stack = inspect.stack()
+    for f in stack:
+        try:
```
stepansergeevitch, line 85: Do you think we can replace this with a for loop?
in `_is_*` functions we basically compare `function` and `filename` to a list of strings. For version import we could've used importlib. And therefore we could basically have a map, which could be extended easily. What do you think?

src/firebolt/utils/usage_tracker.py
```diff
@@ -0,0 +1,144 @@
...
+    return sdk_format + connector_format
+
+
+def get_user_agent_header(
+    custom_connectors: Optional[Union[Tuple[str, str], List[Tuple[str, str]]]] = None
+) -> str:
+    connectors = detect_connectors()
+    logger.debug("Detected running from packages: %s", str(connectors))
+    # Override auto-detected connectors with info provided manually
+    if type(custom_connectors) == tuple:
```
stepansergeevitch, line 139: What if someone would pass
```
(
    ("connector1", "version1"),
    ("connector2", "version2")
)
```

src/firebolt/async_db/connection.py
```diff
@@ -285,17 +291,20 @@ def __init__(
...
-
+        connector_versions = additional_parameters.get("connector_versions", None)
+        user_agent = get_user_agent_header(connector_versions)
         self._client = AsyncClient(
             auth=auth,
             base_url=engine_url,
             api_endpoint=api_endpoint,
             timeout=Timeout(DEFAULT_TIMEOUT_SECONDS, read=None),
             transport=transport,
+            headers={"User-Agent": user_agent},
```
stepansergeevitch, line 307: nit: inline `headers={"User-Agent": get_user_agent_header(connector_versions)}`

stepansergeevitch (approved):
LGTM, left a couple of comments

src/firebolt/utils/usage_tracker.py
stepansergeevitch: Maybe let's just enforce passing a list?

yuryfirebolt (approved):

src/firebolt/utils/usage_tracker.py
ptiurin: Yeah, importlib sounds like a better solution. Initially I thought we might need to use some other properties in a stack frame, but those don't look very useful.

PR #175, opened 2022-06-06
Fir 11539 setting non existing account n

stepansergeevitch:
Fixed missing errors for invalid account_name for `ResourceManager` and `connect`


src/firebolt/async_db/connection.py
```diff
@@ -222,6 +222,16 @@ async def connect_inner(
                 account_name=account_name,
                 api_endpoint=api_endpoint,
             )
+        elif account_name:
+            # We need to manually validate account_name since it's not validated
+            # in neither of above cases
```
ptiurin, line 227: Should it be `if` instead? So if above cases pass it still validates the account name?

stepansergeevitch: Account name gets validated in those cases since it's used to get/resolve engine url.
I'll extend a comment explaining that

PR #179, opened 2022-06-21
Fir 14026 frequently get attributeerror 

stepansergeevitch:
Fixed anyio race condition, which happened when it was lazily importing async backend modules. Now we always import all available backends on the module-level, so this issues doesn't occur anymore. Added integration test

### DIFF
--- a/src/firebolt/client/client.py
+++ b/src/firebolt/client/client.py
@@ -1,5 +1,6 @@
 from typing import Any, Optional
 
+from anyio._core._eventloop import get_asynclib
 from async_property import async_cached_property  # type: ignore
 from httpx import URL
 from httpx import AsyncClient as HttpxAsyncClient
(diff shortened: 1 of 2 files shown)


src/firebolt/client/client.py
```diff
@@ -20,6 +21,15 @@
...
 )
 
+# Explicitly import all available backend not get into
+# anyio race condition during backend import
+for backend in ("asyncio", "trio"):
+    try:
+        get_asynclib(backend)
+    except ModuleNotFoundError:
+        # Not all backends might be installed
+        pass
```
ptiurin, line 31: I'm just worried since we're using a "private" API for this hack it might change without notice. As we're pinning the httpx version we're fine for now, but if we were to upgrade it might cause issues. Hopefully it will be picked up by nightly.

ptiurin (approved):
This is fine as a hack, but I'm worried the functionality might break later on.

src/firebolt/client/client.py
stepansergeevitch: I will investigate how this could be properly fixed

yuryfirebolt (approved):
LGTM

PR #186, opened 2022-07-05
fix: Copy env for nightly

ptiurin:


ptiurin:
Confirmed in nightly this fixes the windows run.

PR #187, opened 2022-07-08
feat: Add ability to do async query execution

ericf-firebolt:
Server-side asynchronous queries now run and return a query_id. Querying of status and retrieval of results is not yet supported. Those will come in a later PR.


tests/unit/db_conftest.py
```diff
@@ -107,6 +107,20 @@ def python_query_data() -> List[List[ColType]]:
     ]
 
 
+@fixture
+def server_side_async_id_callback() -> Response:
+    def do_query(request: Request, **kwargs) -> Response:
+        query_response = {"query_id": "1a3f53d"}
```
ptiurin, line 113: Nit:
```
def server_side_async_id_callback(server_side_async_id) -> Response:
    ...
        query_response = {"query_id": server_side_async_id }
```
So there's a single point where this string is defined.

tests/unit/async_db/test_cursor.py
```diff
@@ -203,8 +255,14 @@ async def test_cursor_execute_error(
...
     for query in (
-        lambda: cursor.execute("select *"),
-        lambda: cursor.executemany("select *", []),
+        (
+            lambda: cursor.execute("select * from t"),
+            "server-side synchronous execute()",
+        ),
+        (
+            lambda: cursor.executemany("select * from t", []),
+            "server-side synchronous executemany()",
```
ptiurin, line 264: We could probably add async counterparts in here as well, since errors should propagate the same way for them.

src/firebolt/async_db/cursor.py
```diff
@@ -327,25 +337,37 @@ async def _api_request(
...
+        async_execution: bool = False,
     ) -> None:
         self._reset()
-        if set_parameters is not None:
-            logger.warning(
-                "Passing set parameters as an argument is deprecated. Please run "
-                "a query 'SET <param> = <value>'"
-            )
+        if async_execution and "use_standard_sql" in self._set_parameters:
+            raise AsyncExecutionUnavailableError()
```
ptiurin, line 364: let's add a test for this as well

src/firebolt/async_db/cursor.py
```diff
@@ -267,6 +271,13 @@ def _reset(self) -> None:
         self._idx = 0
         self._row_sets = []
         self._next_set_idx = 0
+        self._query_id = ""
+
+    def _query_id_from_response_async(self, response: Response) -> str:
+        if response.headers.get("content-length", "") == "0":
+            return ""  # Should this error out?
```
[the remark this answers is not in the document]
ericf-firebolt: I copied `_row_set_from_response()`, and this logic was in there.

src/firebolt/async_db/cursor.py
```diff
@@ -327,25 +337,37 @@ async def _api_request(
...
+            url="/",
+            method="POST",
+            params={
+                "database": self.connection.database,
+                "output_format": JSON_OUTPUT_FORMAT,
+                **self._set_parameters,
+                "async_execution": 1,
+                "advanced_mode": 1,
+            },
+            content=query,
```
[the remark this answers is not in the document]
ericf-firebolt: Ooh, good point.

src/firebolt/async_db/cursor.py
```diff
@@ -327,25 +337,37 @@ async def _api_request(
...
+        )
+
     async def _do_execute(
         self,
         raw_query: str,
         parameters: Sequence[Sequence[ParameterType]],
-        set_parameters: Optional[Dict] = None,
```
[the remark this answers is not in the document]
ericf-firebolt: This was deprecated already, which is why it was removed. @stepansergeevitch ?

tests/unit/async_db/test_cursor.py
```diff
@@ -159,14 +158,24 @@ async def test_cursor_execute(
     """Cursor is able to execute query, all fields are populated properly."""
 
     for query in (
-        lambda: cursor.execute("select * from t"),
-        lambda: cursor.executemany("select * from t", []),
+        (
+            lambda: cursor.execute("select * from t"),
+            "server-side synchronous execute()",
+        ),
```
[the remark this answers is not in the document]
ericf-firebolt: Roger.

tests/unit/db_conftest.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: I had planned to do this in a future PR, but I can do it here if you'd prefer.

src/firebolt/async_db/cursor.py
```diff
@@ -428,23 +454,29 @@ async def execute(
...
         """
         params_list = [parameters] if parameters else []
-        await self._do_execute(query, params_list, set_parameters, skip_parsing)
-        return self.rowcount
+        if async_execution:
+            await self._do_execute(query, params_list, skip_parsing, True)
+            return self.query_id
+        else:
+            await self._do_execute(query, params_list, skip_parsing, False)
+            return self.rowcount
```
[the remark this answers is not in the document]
ericf-firebolt: Much cleaner.

tests/unit/async_db/test_cursor.py
```diff
@@ -175,20 +184,63 @@ async def test_cursor_execute(
...
+    ):
+        # Query with json output
+        httpx_mock.add_callback(auth_callback, url=auth_url)
+        httpx_mock.add_callback(
+            server_side_async_id_callback, url=query_with_params_url
+        )
+
+        assert (
+            await query[0]() == server_side_async_id
+        ), f"Invalid query id returned for {query[1]}."
```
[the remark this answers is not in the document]
ericf-firebolt: Done.

tests/unit/conftest.py
```diff
@@ -155,6 +155,11 @@ def auth_url(settings: Settings) -> str:
     return f"https://{settings.server}{AUTH_URL}"
 
 
+# @fixture
+# def server_side_async_url(settings: Settings) -> str:
```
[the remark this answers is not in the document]
ericf-firebolt: Removed.

src/firebolt/utils/exception.py
```diff
@@ -222,3 +222,17 @@ class NotSupportedError(DatabaseError):
...
+    """
+    If `use_standard_sql` is specified the query status endpoint returns a JSON
+    object with empty values instead of a proper status object. In that case,
+    it is not possible to retrieve the results of an asynchronous query.
+    """
+
+    def __init__(self):  # type: ignore
+        super().__init__(
+            "It is not possible to execute queries asynchronously if "
+            "use_standard_sql is in use."
```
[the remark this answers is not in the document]
ericf-firebolt: Done.

tests/unit/async_db/test_cursor.py
ericf-firebolt: I added them for rowcount and description. Did I miss anything?

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
stepansergeevitch: Having to still support this AND introduce async queries support would complicate the code a lot.
This has been deprecated for like 5 minor versions, and also users can still stick to the old version until they will fix compatibility on their end.

src/firebolt/async_db/cursor.py
```diff
@@ -267,6 +272,13 @@ def _reset(self) -> None:
         self._idx = 0
         self._row_sets = []
         self._next_set_idx = 0
+        self._query_id = ""
+
+    def _query_id_from_response_async(self, response: Response) -> str:
+        if response.headers.get("content-length", "") == "0":
+            raise InternalError("No response to asynchronous query.")
```
stepansergeevitch, line 279: nit: InternalError -> OperationalError

src/firebolt/async_db/cursor.py
```diff
@@ -267,6 +272,13 @@ def _reset(self) -> None:
...
+        query_data = response.json(parse_float=str)
```
stepansergeevitch, line 280: no need for `parse_float` parameter here. It's used for handling decimals in query response. We shouldn't encounter those here

src/firebolt/async_db/cursor.py
```diff
@@ -267,6 +272,13 @@ def _reset(self) -> None:
...
+        return query_data["query_id"]
```
stepansergeevitch, line 281: Validate that `query_id` is in `query_data`

src/firebolt/async_db/cursor.py
```diff
@@ -327,25 +338,31 @@ async def _api_request(
             content=query,
         )
 
+    async def _async_api_request(self, query: str) -> Response:
```
stepansergeevitch, line 341: nit: `_async_execution_api_request`, not to confuse with asyncio

src/firebolt/async_db/cursor.py
```diff
@@ -428,23 +449,28 @@ async def execute(
...
+            async_execution (bool): flag to determine if query should be asynchronous
 
         Returns:
             int: Query row count
         """
         params_list = [parameters] if parameters else []
-        await self._do_execute(query, params_list, set_parameters, skip_parsing)
-        return self.rowcount
+        await self._do_execute(query, params_list, skip_parsing, async_execution)
+        if async_execution:
```
stepansergeevitch, line 462: nit: `return self.query_id if async_execution else self.rowcount`

src/firebolt/async_db/cursor.py
```diff
@@ -327,25 +338,31 @@ async def _api_request(
...
         try:
 
             if parameters and skip_parsing:
                 logger.warning(
                     "Query formatting parameters are provided with skip_parsing."
-                    " They will be ignored"
+                    " They will be ignored."
                 )
 
             # Allow users to manually skip parsing for performance improvement
```
stepansergeevitch, line 368: We also need to error if user tries to execute a multistatement query. We need to error if `len(queries) > 0 and async_execution`

src/firebolt/async_db/cursor.py
```diff
@@ -372,7 +389,7 @@ async def _do_execute(
                     Optional[List[List[RawColType]]],
                 ] = (-1, None, None, None)
                 if isinstance(query, SetParameter):
-                    # Validate parameter by executing simple query with it
+                    # Validate parameter by executing simple query with it.
                     resp = await self._api_request(
```
stepansergeevitch, line 393: Check if user selected `async_execution` here, and warn that since it's a set statement, it will still be executed synchronously

src/firebolt/async_db/cursor.py
```diff
@@ -383,8 +400,12 @@ async def _do_execute(
 
                     # set parameter passed validation
                     self._set_parameters[query.name] = query.value
+                elif async_execution:
+                    resp = await self._async_api_request(query)
+                    await self._raise_if_error(resp)
+                    self._query_id = self._query_id_from_response_async(resp)
                 else:
-                    resp = await self._api_request(query, set_parameters)
+                    resp = await self._api_request(query, {})
```
stepansergeevitch, line 408: nit: make `set_parameters` argument in `_api_request` optional, having None by default

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Roger that.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done. I made it an `OperationalError` if it's missing.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Since you said "warn," I did this via a log statement, but maybe we should actually error out. Later calls to `cancel()` or `get_query_status()` will fail.

tests/unit/async_db/test_cursor.py
ptiurin: I mean that the same cases that are checked in here in `test_cursor_execute_error` can also be applied to the async queries. You can just add those to the list of cases in line258-264 and the same tests would be executed.

src/firebolt/async_db/cursor.py
ptiurin: It's generally a good idea to add tests along with the modified code so I'd prefer to have it here.

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ptiurin: This would mean a breaking change in a minor version though. Should we bump the major instead? Or do we work with the assumption that we might contain breaking changes since we're <1.0?

tests/unit/async_db/test_cursor.py
ericf-firebolt: Actually, I don't think I can add them yet, because `execute(query, async_exectution=True)` just returns a query id number. I can't add async tests until I have the ability to check the query status.

In the meantime what I'll do is add the failure modes below: `OperationalError` if `query_id` is not returned, and two `AsyncExecutionUnavailableError`s, one for when `Set use_standard_sql=1`, and as requested below, one when `len(queries) > 0 and async_execution`. The latter means I'll add another failure mode in `cursor.py`, as well.

src/firebolt/async_db/cursor.py
ericf-firebolt: This is addressed above.

src/firebolt/async_db/cursor.py
ericf-firebolt: Also, a test for this error will also be added, https://github.com/firebolt-db/firebolt-python-sdk/pull/187#discussion_r930235178

src/firebolt/async_db/cursor.py
ericf-firebolt: A test will also be added. https://github.com/firebolt-db/firebolt-python-sdk/pull/187#discussion_r930235178

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ericf-firebolt: Also, added a test, see https://github.com/firebolt-db/firebolt-python-sdk/pull/187#discussion_r930235178

tests/unit/async_db/test_cursor.py
ptiurin: Ah I see, fair point!

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ericf-firebolt: Am I right in that Cosmin ruled against doing this now?

src/firebolt/async_db/cursor.py
ericf-firebolt: Actually, and the return type from all the execute fns will be wrong.

ericf-firebolt: This now errors out and a test is included in `test_cursor.py`.

src/firebolt/async_db/cursor.py
ericf-firebolt: This test is now included. Thanks Stepan!

stepansergeevitch (approved):

PR #189, opened 2022-07-19
fix: pytests now pass with pytest-asyncio v0.19.0

ericf-firebolt:


stepansergeevitch (commented):
I would suggest to have the same naming everywhere
Either do
```
from pytest import fixture
from pytest_asyncio import fixture as async_fixture
...
@fixture
...
@async_fixture
```
Or 
```
import pytest
import pytest_asyncio
...
@pytest.fixture
...
@pytest_asyncio.fixture
```
IMHO, first is a bit cleaner

ptiurin:
Does this work for older pythons as well? e.g. `3.*.1` - `3.*.10`

ericf-firebolt:
@stepansergeevitch, that was an oversight. I had intended it to be 
```import pytest
import pytest_asyncio
...
@pytest.fixture
...
@pytest_asyncio.fixture```
wherever it occurs, but I seem to have missed this module. You're right that the second seems clearer, though. I'll go through and change it wherever it occurs.

@ptiurin , I need to test for that. To be clear, I should test in, for instance, 3.7.10, 3.8.10, 3.9.10? I remember there was an issue with Python versions like 3.9.12, where the minor version was > 10. Is that what you're asking about?

ericf-firebolt:
> I would suggest to have the same naming everywhere Either do
> 
> ```
> from pytest import fixture
> from pytest_asyncio import fixture as async_fixture
> ...
> @fixture
> ...
> @async_fixture
> ```
> 
> Or
> 
> ```
> import pytest
> import pytest_asyncio
> ...
> @pytest.fixture
> ...
> @pytest_asyncio.fixture
> ```
> 
> IMHO, first is a bit cleaner

@stepansergeevitch, that was an oversight. I had intended it to be 
```import pytest
import pytest_asyncio
...
@pytest.fixture
...
@pytest_asyncio.fixture```
wherever it occurs, but I seem to have missed this module. You're right that the first seems clearer, though. I'll go through and change it wherever it occurs.

ericf-firebolt:
> Does this work for older pythons as well? e.g. `3.*.1` - `3.*.10`

It passes our PR unit tests, which use Python 3.7.13, so we should be okay.

ptiurin:
> > Does this work for older pythons as well? e.g. `3.*.1` - `3.*.10`
> 
> It passes our PR unit tests, which use Python 3.7.13, so we should be okay.

Yeah, but does it pass on patch version <.11? e.g. 3.7.10? That's what I'm worried about, if this is backwards-compatible.

ericf-firebolt:
> > > Does this work for older pythons as well? e.g. `3.*.1` - `3.*.10`
> > 
> > 
> > It passes our PR unit tests, which use Python 3.7.13, so we should be okay.
> 
> Yeah, but does it pass on patch version <.11? e.g. 3.7.10? That's what I'm worried about, if this is backwards-compatible.

Oh, got it: all dogs are mammals, but not all mammals are dogs.

All unit tests pass on Python 3.7.10 and 3.9.10. Integration tests pass on github (although  not locally for me). I think we're clear to merge.

Note that I also pinned pytest-asyncio to v0.19.0 in `setup.cfg`. Not sure this is a good idea, but I leave it to your judgement.

ericf-firebolt:
> 

Okay, I fixed the one I missed. I think things are good to go now.

ericf-firebolt:
> I would suggest to have the same naming everywhere Either do
> 
> ```
> from pytest import fixture
> from pytest_asyncio import fixture as async_fixture
> ...
> @fixture
> ...
> @async_fixture
> ```
> 
> Or
> 
> ```
> import pytest
> import pytest_asyncio
> ...
> @pytest.fixture
> ...
> @pytest_asyncio.fixture
> ```
> 
> IMHO, first is a bit cleaner

Okay, I fixed the one I missed. I think things are good to go now.

stepansergeevitch (approved):
LGTM, please run integration tests action for this pr before merging

PR #192, opened 2022-08-02
feat: Add cancel and status queries to server-side async execution

ericf-firebolt:


src/firebolt/async_db/cursor.py
```diff
@@ -61,6 +61,15 @@ class CursorState(Enum):
     CLOSED = 4
 
 
+class QueryStatus(Enum):
+    """Class for query responses on server-side async queries."""
+
+    RUNNING = 1
+    ENDED_SUCCESSFULLY = 2
+    ENDED_UNSUCCESSFULLY = 3
+    NOT_AVAILABLE = 4
```
stepansergeevitch, line 70: nit: Rename to NOT_READY. NOT_AVAILABLE might give users the impression that it's will never be available, but the case is it's just not ready yet.

src/firebolt/async_db/cursor.py
```diff
@@ -419,9 +419,16 @@ async def _do_execute(
                 if isinstance(query, SetParameter):
                     await self._validate_set_parameter(query)
                 elif async_execution:
-                    resp = await self._async_execution_api_request(query)
-                    await self._raise_if_error(resp)
-                    self._query_id = self._query_id_from_response_async(resp)
+                    response = await self._async_execution_api_request(query)
+                    await self._raise_if_error(response)
+                    if response.headers.get("content-length", "") == "0":
```
stepansergeevitch, line 424: Why did you have to inline this part?

src/firebolt/async_db/cursor.py
```diff
@@ -328,17 +327,18 @@ def _append_row_set(
 
     async def _api_request(
         self,
-        query: str,
-        set_parameters: Optional[dict] = None,
+        query: Optional[str] = "",
+        parameters: Optional[dict] = None,
+        url_suffix: Optional[str] = "",
```
stepansergeevitch, line 332: nit: url_suffix -> path

src/firebolt/async_db/cursor.py
```diff
@@ -584,6 +591,49 @@ def setinputsizes(self, sizes: List[int]) -> None:
     def setoutputsize(self, size: int, column: Optional[int] = None) -> None:
         """Set a column buffer size for fetches of large columns (does nothing)."""
 
+    @check_not_closed
+    async def cancel(self, query_id: str) -> None:
+        """Cancel a server-side async query."""
+        try:
+            resp = await self._api_request(
+                parameters={"query_id": query_id}, url_suffix="cancel"
+            )
```
stepansergeevitch, line 600: add `self._raise_if_error`. Also check if all it's functionality fits here

src/firebolt/async_db/cursor.py
```diff
@@ -584,6 +591,49 @@ def setinputsizes(self, sizes: List[int]) -> None:
...
+            raise
+
+    @check_not_closed
+    async def get_status(self, query_id: str) -> QueryStatus:
+        """Get status of a server-side async query. Return the state of the query."""
+        try:
+            resp = await self._api_request(
+                parameters={"query_id": query_id}, url_suffix="status"
+            )
+            # if resp.status_code == codes.BAD_REQUEST:
```
stepansergeevitch, line 616: remove comment

src/firebolt/async_db/cursor.py
```diff
@@ -584,6 +591,49 @@ def setinputsizes(self, sizes: List[int]) -> None:
...
+            #         f"Asynchronous query {query_id} status check failed."
+            #     )
+            if resp.headers.get("content-length", "") == "0":
+                raise OperationalError("No response to asynchronous query.")
+            if resp.status_code == codes.BAD_REQUEST:
+                return QueryStatus.NOT_AVAILABLE
+            resp_json = resp.json()
+            if "status" not in resp_json:
+                raise OperationalError(
+                    f"Invalid response to asynchronous query: missing query ID."
```
stepansergeevitch, line 627: It probably should be "missing status" or something

src/firebolt/async_db/cursor.py
```diff
@@ -584,6 +591,49 @@ def setinputsizes(self, sizes: List[int]) -> None:
...
+        """Get status of a server-side async query. Return the state of the query."""
+        try:
+            resp = await self._api_request(
+                parameters={"query_id": query_id}, url_suffix="status"
+            )
+            # if resp.status_code == codes.BAD_REQUEST:
+            #     raise OperationalError(
+            #         f"Asynchronous query {query_id} status check failed."
+            #     )
+            if resp.headers.get("content-length", "") == "0":
```
stepansergeevitch, line 620: I think this case will be caught with response.json

tests/integration/dbapi/async/test_queries_async.py
```diff
@@ -353,3 +355,46 @@ async def test_set_invalid_parameter(connection: Connection):
...
 
         assert len(c._set_parameters) == 0
+
+
+@mark.asyncio
+async def test_ss_async_execution_query(connection: Connection) -> None:
+    """Make an sql query and receive an id back."""
+    with connection.cursor() as c:
+        query_id = await c.execute(
+            "DROP TABLE IF EXISTS test_tb", [], async_execution=True
```
stepansergeevitch, line 365: Make sure test_db exists

tests/integration/dbapi/async/test_queries_async.py
```diff
@@ -353,3 +355,46 @@ async def test_set_invalid_parameter(connection: Connection):
...
+        )
+    assert type(query_id) is str and query_id
+
+
+@mark.asyncio
+async def test_ss_async_execution_cancel(connection: Connection) -> None:
+    """Test cancel."""
+    with connection.cursor() as c:
+        query_id = await c.execute(
+            "DROP TABLE IF EXISTS test_tb", [], async_execution=True
```
stepansergeevitch, line 375: Try simulating the case where you cancel a query and validate that it was actually canceled.
I can think of something like
```
SET use_standard_sql=0
INSERT INTO tb SELECT sleepEachRow(1) FROM numbers(<seconds>)
```
Choose seconds to be not too much not to increase integration tests execution time, but enough to be able to see that it was canceled.

stepansergeevitch: And after it's canceled check that not all values were inserted (it was canceled during execution)

src/firebolt/async_db/cursor.py
```diff
@@ -328,17 +331,35 @@ def _append_row_set(
...
+                    must be an empty string. If no value not specified,
+                    JSON_OUTPUT_FORMAT will be used.
+            url_suffix (str): endpoint suffix, for example "cancel" or "status"
+        """
+        if not parameters:
+            parameters = {"output_format": JSON_OUTPUT_FORMAT}
+        elif "output_format" not in parameters:
+            parameters["output_format"] = JSON_OUTPUT_FORMAT
+        elif not parameters["output_format"]:  # if "output_format" is empty string
+            del parameters["output_format"]
```
ptiurin, line 355: When would this happen?

src/firebolt/async_db/cursor.py
```diff
@@ -366,7 +387,7 @@ async def _validate_set_parameter(self, parameter: SetParameter) -> None:
         # set parameter passed validation
         self._set_parameters[parameter.name] = parameter.value
 
-    async def _do_execute(
+    async def _do_execute(  # noqa: C901
```
ptiurin, line 390: I think we shouldn't just blanket ignore warnings when they pop up. If a function is becoming complex we'd better refactor it. SonarCloud is also triggering on this, but not failing.

src/firebolt/async_db/cursor.py
```diff
@@ -594,6 +665,9 @@ def __exit__(
     ) -> None:
         self.close()
 
+    # @check_not_closed
+    # def fetch_async_query_result(query_id: str):
```
ptiurin, line 669: comment

src/firebolt/async_db/cursor.py
```diff
@@ -584,6 +612,49 @@ def setinputsizes(self, sizes: List[int]) -> None:
...
+            self._state = CursorState.ERROR
+            raise
+
+    @check_not_closed
+    async def get_status(self, query_id: str) -> QueryStatus:
+        """Get status of a server-side async query. Return the state of the query."""
+        try:
+            resp = await self._api_request(
+                # output_format must be empty for status to work correctly.
+                parameters={"query_id": query_id, "output_format": ""},
```
ptiurin, line 636: Since we're deleting the key "output_format" in `_api_request`, wouldn't it make sense to not provide it in the first place?

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: Okay. I was thinking that `resp.json()` might not exist and we'd get a `KeyError`.

tests/integration/dbapi/async/test_queries_async.py
ericf-firebolt: I changed it to `SELECT 1`. All I care about is getting a query_id back.

src/firebolt/async_db/cursor.py
ericf-firebolt: In order to make sure that `output_format` isn't always set to `JSON_OUTPUT_FORMAT` we may occasionally need to send in an empty value. In that case, it needs to be removed or our API endpoint string will be malformed.

src/firebolt/async_db/cursor.py
ericf-firebolt: Darn it.

src/firebolt/async_db/cursor.py
ericf-firebolt: That is kind of a weird functionality and confusing. Alternately, I could add a parameter to the function signature for the `output_format` value.

src/firebolt/async_db/cursor.py
ericf-firebolt: The thing is that it has a default value, so if it's missing the default needs to be subbed in. I agree that this isn't clean though. I'll give this some thought.

src/firebolt/async_db/cursor.py
ericf-firebolt: I think that might have been cruft from an earlier design decision that was changed. I removed it.

src/firebolt/async_db/cursor.py
ericf-firebolt: I think you're right. I removed this check. Instead, after a cancel the user can check the status of the query.

src/firebolt/async_db/cursor.py
ptiurin: Ah I get it now. Since `_api_request` is an internal function, can we add a check for no output format from user input on outside of it, in whichever function calls it? Then when we need to call `_api_request` with no output format we can call it as is.

tests/integration/dbapi/async/test_queries_async.py
ericf-firebolt: I couldn't get `SET use_standard_sql=0; INSERT INTO tb SELECT sleepEachRow(1) FROM numbers(<seconds>)` to work with server-side async queries. It failed saying ss-async can't be used with multi-statement queries. And my attempts to first `SET use_standard_sql=0` in an earlier call to `execute()` also failed. Instead, I just created a long insert. At a certain point this will fail, as Firebolt gets better at inserting, but for now it's the best I can do.

src/firebolt/async_db/cursor.py
ericf-firebolt: There actually never was a test before. It was always `JSONCompact`, which is what caused the problem. I added this functionality only for `get_status()`, which is why this solution isn't great. 

I also, just now, went ahead and changed `cancel()` so it removes it, and also when validating a `SET` parameter.

src/firebolt/async_db/cursor.py
ericf-firebolt: Done.

src/firebolt/async_db/cursor.py
ericf-firebolt: See further discussion: https://github.com/firebolt-db/firebolt-python-sdk/pull/192#discussion_r941045329

src/firebolt/async_db/cursor.py
ericf-firebolt: This has been resolved.

src/firebolt/async_db/cursor.py
ericf-firebolt: This was dealt with. The user does not have the option to choose output formats.

src/firebolt/async_db/cursor.py
```diff
@@ -366,6 +387,35 @@ async def _validate_set_parameter(self, parameter: SetParameter) -> None:
...
         self._set_parameters[parameter.name] = parameter.value
 
+    def _validate_ss_async_settings(
+        self,
+        parameters: Sequence[Sequence[ParameterType]],
+        queries: List[Union[SetParameter, str]],
+        skip_parsing: bool = False,
+        async_execution: Optional[bool] = False,
+    ) -> None:
+        if async_execution and self._set_parameters.get("use_standard_sql", "1") == "0":
```
aymeric-dispa, line 397: Do we need to do a validation here ? (isn't it validated by backend?)

src/firebolt/async_db/cursor.py
```diff
@@ -374,34 +424,14 @@ async def _do_execute(
...
-                    "It is not possible to execute multi-statement "
-                    "queries asynchronously."
-                )
             for query in queries:
 
                 start_time = time.time()
-                # our CREATE EXTERNAL TABLE queries currently require credentials,
+                # Our CREATE EXTERNAL TABLE queries currently require credentials,
                 # so we will skip logging those queries.
                 # https://docs.firebolt.io/sql-reference/commands/ddl-commands#create-external-table
```
aymeric-dispa, line 436: I think it was moved to https://docs.firebolt.io/sql-reference/commands/create-external-table.html

src/firebolt/async_db/cursor.py
ericf-firebolt: I don't think so. I was having no issues setting these to incompatible values. In fact, it was making my debugging incredibly annoying.

Do you deal with this in the JDBC?

aymeric-dispa: I don't validate specific params with JDBC (once a param is set I call SELECT 1;). If it works, fine, if not then I return the error from backend .

src/firebolt/async_db/cursor.py
ptiurin: @ericf-firebolt This is what I meant by moving it outside, https://github.com/firebolt-db/firebolt-python-sdk/pull/196

ericf-firebolt: Ah, got it. Yeah, that's much cleaner. I feel like there's a reason I didn't to it, but it was a mistake not to. Anyway, that's merged.

src/firebolt/async_db/cursor.py
ericf-firebolt: It doesn't actually return an error. Instead, the query id never gets written to the query_history table so we just get query not ready messages. I don't see a way around this other than checking in the SDK. I'm going to resolve this.

docsrc/Connecting_and_queries.rst
```diff
@@ -105,35 +105,35 @@ To get started, follow the steps below:
...
 
-	The ``cursor`` object can be used to send queries and commands to your Firebolt database and engine. See below for examples of functions using the ``cursor`` object. 
+	The ``cursor`` object can be used to send queries and commands to your Firebolt database and engine. See below for examples of functions using the ``cursor`` object.
 
-Command and query examples
+Server-side synchronous command and query examples
 ============================
 
-This section includes Python examples of various SQL commands and queries. 
+This section includes Python examples of various SQL commands and queries. For running queries in server-side asynchronous mode see :ref:`server-side asynchronous query execution`.
```
kevinmarr, line 113: What's the reasoning for adding this? Seems random to call out one feature of all the features listed in this section.

docsrc/Connecting_and_queries.rst
```diff
@@ -246,32 +246,110 @@ Multiple-statement queries allow you to run a series of SQL statements sequentia
...
+.. note::
+
+	Multiple statement queries are not able to use placeholder values for parameterized queries.
+
+
+
+Server-side asynchronous query execution
+==========================================
+
+Server-side asynchronous query execution allows you to run a long query in the background while executing other asynchronous or synchronous queries. It is not possible to retrieve the results of a server-side asynchronous query, so these queries are best used for running lambda functions, or to run SELECTs for warming the cache.
```
kevinmarr, line 265: When I mentioned Lambda I was referring to AWS Lambda, which I don't think is the same thing as lambda functions necessarily. Maybe we should just generically say "DMLs" instead of lambda functions? Also, maybe it would be good to explicitly say that the benefit of server-side async is to be able to free up a connection or potentially even spin down an entire service (such as AWS Lambda) while a long-running database job is still underway.

docsrc/Connecting_and_queries.rst
ericf-firebolt: Because server-side queries are a whole other section, not just one more feature, but I'm not wedded to this. I did mean to write something like "this is for server-side synchronous..."

docsrc/Connecting_and_queries.rst
ericf-firebolt: Oh, thanks for the clarity. I was going off memory of our conversation and that had begun to tread into areas which I don't have any real experience with, hence not understanding Lambda vs lambda (which I didn't think we provided).

src/firebolt/async_db/cursor.py
ericf-firebolt: Got it.

docsrc/Connecting_and_queries.rst
ericf-firebolt: I did this and basically quoted you.

src/firebolt/async_db/cursor.py
```diff
@@ -366,6 +377,35 @@ async def _validate_set_parameter(self, parameter: SetParameter) -> None:
...
+                "It is not possible to execute queries asynchronously if "
+                "use_standard_sql=0."
+            )
+        if parameters and skip_parsing:
+            logger.warning(
+                "Query formatting parameters are provided but skip_parsing "
+                "is specified. They will be ignored."
+            )
+
+        # Allow users to manually skip parsing for performance improvement.
```
ptiurin, line 398: I think this comment got lost?

tests/integration/dbapi/async/test_queries_async.py
```diff
@@ -341,6 +409,77 @@ async def test_set_invalid_parameter(connection: Connection):
...
+
+async def test_server_side_async_execution_get_status(connection: Connection) -> None:
+    """
+    Test get_status(). Test for three ending conditions: PARSE_ERROR,
+    STARTED_EXECUTION, ENDED_EXECUTION.
+    """
+    with connection.cursor() as c:
+        try:
+            await c.execute(
+                "CREATE DIMENSION TABLE IF NOT EXISTS test_tbl (id int, name string)"
```
ptiurin, line 457: This can also be a global. In fact, since this as well as `LONG_INSERT` are reused in both sync and async tests they can be placed at the dbapi/ conftest level.

ptiurin (approved):
Minor comments, otherwise lgtm.

docsrc/firebolt.async_db.rst
```diff
@@ -2,7 +2,7 @@
 Async DB
 ==========================
 
-The Async DB package enables connecting to a Firebolt database for asynchronous queries.
+The Async DB package enables connecting to a Firebolt database for `client-side` asynchronous queries. For running queries in `server-side` asynchronous mode see :ref:`server-side asynchronous query execution`.
```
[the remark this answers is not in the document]
ericf-firebolt: Good point, but I think that mention should be on the main page for async queries: connecting_and_queries.rst, which is where the other benefits of server-side async queries are listed.

src/firebolt/async_db/cursor.py
```diff
@@ -328,27 +331,35 @@ def _append_row_set(
...
+            query (str): SQL query
+            parameters (Optional[Sequence[ParameterType]]): A sequence of substitution
+                parameters. Used to replace '?' placeholders inside a query with
+                actual values. Note: In order to "output_format" dict value, it
+                    must be an empty string. If no value not specified,
+                    JSON_OUTPUT_FORMAT will be used.
+            path (str): endpoint suffix, for example "cancel" or "status"
+        """
+        if use_set_parameters:
+            parameters = {**(self._set_parameters or {}), **(parameters or {})}
```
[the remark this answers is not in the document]
ericf-firebolt: I think sphinx will eventually complain about it, but it would be swamped by other warnings and would probably be missed.

src/firebolt/async_db/cursor.py
```diff
@@ -366,6 +377,35 @@ async def _validate_set_parameter(self, parameter: SetParameter) -> None:
         # set parameter passed validation
         self._set_parameters[parameter.name] = parameter.value
 
+    def _validate_ss_async_settings(
```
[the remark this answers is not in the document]
ericf-firebolt: This was an oversight. It's been renamed to match the other function names, which all spell out `server_side`.

src/firebolt/async_db/cursor.py
ericf-firebolt: Actually, this entire function is an orphan. I killed it.

ericf-firebolt: Oops!

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ericf-firebolt: I _think_ sphinx would catch this, but in any case, done.

tests/integration/dbapi/async/test_queries_async.py
ericf-firebolt: I made this and the `DROP TABLE` below both globals, but I don't know how to move them out of their modules and up a level: I couldn't get the imports to work, and  pytest won't import anything other than fixtures from the conftest files, as far as I can tell.

docsrc/Connecting_and_queries.rst
```diff
@@ -246,32 +246,110 @@ Multiple-statement queries allow you to run a series of SQL statements sequentia
...
+
+	Multiple statement queries are not able to use placeholder values for parameterized queries.
+
+
+
+Server-side asynchronous query execution
+==========================================
+
+Server-side asynchronous query execution allows you to run a long query in the background while executing other asynchronous or synchronous queries. An additional benefit of server-side async execution that can free up a connection or potentially even spin down an entire service (such as AWS Lambda) while a long-running database job is still underway. Note that it is not possible to retrieve the results of a server-side asynchronous query, so these queries are best used for running DMLs and DDLs. SELECTs should be used only for warming the cache.
+
```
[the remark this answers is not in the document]
kevinmarr: @ericf-firebolt I see you merged the PR, but what about this comment above?

docsrc/Connecting_and_queries.rst
```diff
@@ -246,32 +246,110 @@ Multiple-statement queries allow you to run a series of SQL statements sequentia
...
+Running DDL commands
+-----------------------------
+
+.. _basic_execute_example:
+
+Running queries server-side asynchronously is similar to running server-side asynchronous queries, but the ``execute()`` command receives an extra parameter, ``async_execution=True``. The example below uses ``cursor`` to create a new table called ``test_table``. ``execute(query, async_execution=True)`` will return a query ID, which can subsequently be used to check the query status.
```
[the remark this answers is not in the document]
kevinmarr: @ericf-firebolt I see you merged the PR, but what about this comment above?

docsrc/Connecting_and_queries.rst
```diff
@@ -246,32 +246,110 @@ Multiple-statement queries allow you to run a series of SQL statements sequentia
...
+    * ``ENDED_SUCCESSFULLY``
+    * ``ENDED_UNSUCCESSFULLY``
+    * ``NOT_READY``
+    * ``STARTED_EXECUTION``
+    * ``PARSE_ERROR``
+    * ``CANCELED_EXECUTION``
+    * ``EXECUTION_ERROR``
+
+
+Once the status of the table creation is ``ENDED_SUCCESSFULLY`` created, data can be inserted into it:
```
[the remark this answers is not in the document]
kevinmarr: @ericf-firebolt I see you merged the PR, but what about this comment above?

tests/integration/dbapi/async/test_queries_async.py
ericf-firebolt: Actually, I had to reopen this. Can you tell me how to get the import to work? It seems to be beyond my Python abilities.

kevinmarr (commented):
not sure how GH reviews work. do i need to submit my review for you to see my comments? testing testing...

PR #195, opened 2022-08-16
feat: Update interface error messages

ericf-firebolt:
This is just a draft so you all can comment on my current solution.


src/firebolt/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.9.2"
+__version__ = "0.9.3"
```
ptiurin, line 1: This does not need to be updated manually, we have a GH action that does it.

ptiurin (approved):
Apart from that looks good.

src/firebolt/__init__.py
ericf-firebolt: Yeah, I mentioned this in Slack https://firebolt-analytics.slack.com/archives/C03748V9JK0/p1660675364408029. As a short-term solution, I had to make dbt install the SDK from a specific branch on github. The only way I could be certain this was happenning correctly was to bump the version number. I'll decrement it again when it's ready to merge.

Which, it's not because I'm still battling the unit tests.

tests/unit/conftest.py
```diff
@@ -197,25 +198,87 @@ def do_mock(
...
 
 @fixture
-def get_engine_callback(
-    get_engine_url: str, engine_id: str, settings: Settings
+def get_engine_id_by_name_url(
+    settings: Settings, account_id: str, engine_name: str
+) -> str:
+    return f"https://{settings.server}" + ACCOUNT_ENGINE_ID_BY_NAME_URL.format(
+        account_id=account_id
+    )
```
ptiurin, line 227: You've can use the `engine_name` parameter here since the URL hit is actually `https://api.mock.firebolt.io/core/v1/accounts/mock_account_id/engines:getIdByName?engine_name=mock_engine_name`

e.g.
```
    return f"https://{settings.server}" + ACCOUNT_ENGINE_ID_BY_NAME_URL.format(
        account_id=account_id
    ) + f"?engine_name={engine_name}"
```

This way you don't need to do anything special in the actual test itself.

tests/unit/conftest.py
```diff
@@ -197,25 +198,87 @@ def do_mock(
...
+    def do_mock(
+        request: Request = None,
+        **kwargs,
+    ) -> Response:
+        assert request.url == get_engine_id_by_name_url
+        return Response(
+            status_code=httpx.codes.OK,
+            json={
+                "engine_id": {
+                    "engine_id": "engine_id",
```
ptiurin, line 252: This is not supposed to be a string, but rather `engine_id` fixture. 
```
"engine_id": engine_id,
```

tests/unit/conftest.py
ericf-firebolt: Actually, it turns out that this and `get_engine_id_by_name_callback()` aren't being used. It's odd, because I expected them to be necessary in `test_connect_engine_name()`, but they're not. So I just deleted them.

ptiurin: They're not used because you're doing it in the test itself, https://github.com/firebolt-db/firebolt-python-sdk/pull/195/files#diff-73d034cb3e24eab31a63fa270a1c5a0b52ab2f6784094e8022b57c8d8f2472deR193
You can either have it as `add_callback` and use fixtures or have `add_response` as it is now.

ericf-firebolt: omfg. I just merged this damn thing.

I'll open a new PR.

PR #196, opened 2022-08-18
Moving JSON_OUTPUT_FORMAT outside of _api_request

ptiurin:


ericf-firebolt (approved):
Yeah, this is much cleaner. Don't remember why I decided against it when I made the changes. 🤷‍♀️

PR #197, opened 2022-08-18
feat: add get_default_engine method for the database

yuryfirebolt:


tests/integration/resource_manager/test_database.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    rm = ResourceManager(rm_settings)
+
+    db = rm.databases.get_by_name(database_name)
+
+    engine = db.get_default_engine()
+    assert engine is not None, "default engine is None, but shouldn't"
+    assert engine.name in [
+        stopped_engine_name,
+        engine_name,
```
ptiurin, line 22: We can safely assume `engine_name` is the default one. It's always created first. This way this test has more substance as well.

ptiurin (approved):
Apart from that lgtm

ericf-firebolt (approved):
lgtm

PR #199, opened 2022-08-24
fix: Several changes were missed when Async queries add cancel status was first merged

ericf-firebolt:
There were a few uncommitted changes when I merged this branch previously. Sigh.


ericf-firebolt (commented):
n/a

tests/integration/dbapi/async/test_queries_async.py
```diff
@@ -8,6 +8,13 @@
 from firebolt.async_db._types import ColType, Column
 from firebolt.async_db.cursor import QueryStatus
 
+VALS_TO_INSERT = ",".join([f"({i},'{val}')" for (i, val) in enumerate(range(1, 360))])
```
ericf-firebolt, line 11: @ptiurin , I don't know how to make these constants global and move them out of this module to the parent directory. Can you give me some guidance?

ptiurin: @ericf-firebolt , you can place them in conftest file, similar to https://github.com/firebolt-db/firebolt-python-sdk/blob/main/tests/integration/conftest.py#L19
Since conftest is picked up on pytest run it loads fixtures from there as well as any constants.

ptiurin: or rather down a level in https://github.com/firebolt-db/firebolt-python-sdk/blob/main/tests/integration/dbapi/conftest.py#L19

ericf-firebolt: When I tried that it didn't work. It looked to me like pytest picks up @fixtures, but that picking up a global is a Python import issue. I'll try again in a sec, though. 

I also tried by creating an `__init__.py` file and by creating a `globals.py` file, but couldn't get the relative imports to work. I'll will admit here that I find `__init.py__` files and Python imports to be a very dark art.

ericf-firebolt: No, the variable in that `conftest.py` are only used locally to the module. What I did instead was create a setup/teardown fixture in `conftest.py`. Annoyingly, I had to create two, one for async, because you can't 
`c.execute('CREATE yadayada')`
`query_id = await c.execute('INSERT INTO yadayada values (yada1, yada2)', async_execute=True)`
If you try you get an error about stuff not being awaited. You have to await the first `execute()`, as well.

ptiurin: Ah, you're right. Those examples only work if there's a fixture that returns the sql string.

kevinmarr (approved):
lgtm

kevinmarr:
@ericf-firebolt once merged, make sure to make a new build of our readthedocs site. ask @FireboltCorey if you have questions on how to do this.

tests/integration/dbapi/sync/test_queries.py
```diff
@@ -1,12 +1,27 @@
...
+    OperationalError,
+    connect,
+)
+
+VALS_TO_INSERT = ",".join([f"({i},'{val}')" for (i, val) in enumerate(range(1, 360))])
+LONG_INSERT = f"INSERT INTO test_tbl VALUES {VALS_TO_INSERT}"
+CREATE_TEST_TABLE = (
+    "CREATE DIMENSION TABLE IF NOT EXISTS test_tbl (id int, name string)"
+)
+DROP_TEST_TABLE = "DROP TABLE IF EXISTS test_tbl"
```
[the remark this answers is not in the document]
ericf-firebolt: You beat me to it.

PR #204, opened 2022-09-08
fix: Engine is running check

ptiurin:
This was creating an issue when an incorrect query was executed.

Two things were the issue here:
- there's a new engine status
- engine url has dashes where name has underscores


stepansergeevitch (approved):

ptiurin:
Confirmed this is the case for both prod and dev now. Ready to merge.

PR #205, opened 2022-09-21
fix: Is db available check

ptiurin:
API has changed in dev. If db is not available we get a 404.

NOTE: This does not work on prod so if we merge it users might see a problem.


ptiurin:
On hold until prod is in line with dev to avoid impact on customers.

ptiurin:
This still differs dev vs. prod

ptiurin:
Closing stale PR. We can redo it afresh if needed.

PR #208, opened 2022-09-27
test: Action cli

ptiurin:


ArsenijKroptya (commented):
Do we on purpose store executable dev_cli in the git repo instead of some artifacts storage?

PR #220, opened 2022-11-23
feat: use trio for async_to_sync

ptiurin:
Asyncio prevents us from sharing Connection objects across threads due to registering httpx Client to the main thread's loop.  Trio seems to solve this problem, as confirmed in the integration test I've added. Looks like there's no impact on the rest of the functionality.

### DIFF
--- a/src/firebolt/utils/util.py
+++ b/src/firebolt/utils/util.py
@@ -4,7 +4,7 @@ from asyncio import (
     new_event_loop,
     set_event_loop,
 )
-from functools import lru_cache, wraps
+from functools import lru_cache, partial, wraps
 from threading import Thread
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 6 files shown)


src/firebolt/utils/util.py
```diff
@@ -84,100 +70,19 @@ def fix_url_schema(url: str) -> str:
...
-            loop = new_event_loop()
-            set_event_loop(loop)
-        # We are inside a running loop
-        if loop.is_running():
-            nonlocal async_job_thread
-            if not async_job_thread:
-                async_job_thread = AsyncJobThread()
-            return async_job_thread.execute(f(*args, **kwargs))
-        return loop.run_until_complete(f(*args, **kwargs))
+        return trio.run(partial(f, *args, **kwargs))
```
stepansergeevitch, line 85: nice

stepansergeevitch (approved):
Looks good and much simpler. Can you please also try to run async queries in jupyter, I believe it was flaky as some point

PR #228, opened 2023-01-12
test: Fix integration tests

stepansergeevitch:
Made a couple of changes in order to fix integration tests
- removed outdated `firebolt_use_decimal` flag. This seems to be the default behavior now
- remove `select 1` from expected to fail statements for system engine. System engine seems to support this statements now
The last integration tests run was successful: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/3901616576

### DIFF
--- a/tests/integration/dbapi/sync/test_system_engine.py
+++ b/tests/integration/dbapi/sync/test_system_engine.py
@@ -137,3 +137,10 @@ def test_start_stop_engine(setup_dbs, connection_system_engine, engine_name):
         check_engine_status(cursor, engine_name, "Running")
         cursor.execute(f"STOP ENGINE {engine_name}")
         check_engine_status(cursor, engine_name, "Stopped")
+
+
+@mark.xdist_group(name="system_engine")
+def test_select_one(connection_system_engine):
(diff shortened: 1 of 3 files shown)


tests/integration/dbapi/sync/test_system_engine.py
```diff
@@ -68,7 +68,7 @@ def db_specs(region, attached_engine):
 
 @mark.parametrize(
     "query",
-    ["SELECT 1", "CREATE DIMENSION TABLE dummy(id INT)", "SHOW TABLES", "SHOW INDEXES"],
+    ["CREATE DIMENSION TABLE dummy(id INT)", "SHOW TABLES", "SHOW INDEXES"],
```
ptiurin, line 71: Should we add a test for SELECT 1 to succeed since that's the expected behaviour now?

ptiurin (approved):
Otherwise, LGTM

tests/integration/dbapi/sync/test_system_engine.py
stepansergeevitch: Makes sense

PR #229, opened 2023-01-16
feat: Fir 18774 new date types

stepansergeevitch:
Added support for new datatypes: `PGDate`, `TimestampNtz` and `TimestampTz`. This also required to extend the range of supported datetime incoming strings. The timezone suffix can now be provided in following the format `[+-]\d\d[:\d\d[:\d\d]]`.
Unfortunately, python `datetime` module (up to python 3.10) doesn't support timezones in a form `[+-]\d\d`, so we need to fix the datetime string for this case (otherwise parsing fails). We also need to fix a datetime string in case if milliseconds digits count doesn't equal to 3 and 6 since `datetime` module doesn't support those as well.
After these changes, `ciso8601` doesn't cover the full range of supported datetime strings anymore, so we might consider dropping it's support if future. As an alternative, we can consider using a backport of python 3.11 new datetime implementation. This one covers all of our cases, as well as has the same speed as `ciso8601`. Yet, it required to be build when installed, which might make the sdk installation process a bit tricky for users.


src/firebolt/async_db/_types.py
```diff
@@ -18,13 +19,26 @@
...
-        # fromisoformat doesn't support milliseconds
-        if "." in date_string:
-            return datetime.strptime(date_string, format)
-        return datetime.fromisoformat(date_string)
+    unsupported_milliseconds_re = re.compile(r"(?<=\.)\d{1,5}(?!\d)")
+
+    def _fix_milliseconds(datetime_string: str) -> str:
+        # Fill milliseconds with 0 to have exactly 6 digits
+        # Python parser only supports 3 or 6 digit milliseconds untill 3.11
+        def allign_ms(match: re.Match) -> str:
```
ptiurin, line 27: nit: `align`

ptiurin (approved):
Otherwise lgtm

Issue #227, opened 2022-12-28
Relaxing `pydantic[dotenv]` bounds


zzstoatzz:
Hello!

I'm writing integrations for [Prefect](https://github.com/PrefectHQ/prefect) on top of this sdk and running into issues with the enforced dependencies' versions.

Specifically, what I'm working on requires `pydantic >= 1.10` so I'm running into this when using `firebolt-sdk==0.13.1`
```
ERROR: Cannot install prefect, pydantic[dotenv]==1.8.2, pydantic[dotenv]==1.9.0, pydantic[dotenv]==1.9.1 and pydantic[dotenv]==1.9.2 because these package versions have conflicting dependencies.

The conflict is caused by:
    prefect 2.7.0 depends on pydantic>=1.10.0
    fastapi 0.70.0 depends on pydantic!=1.7, !=1.7.1, !=1.7.2, !=1.7.3, !=1.8, !=1.8.1, <2.0.0 and >=1.6.2
    pydantic[dotenv] 1.9.2 depends on pydantic 1.9.2
```

and if I try to use an earlier version, e.g.`firebolt-sdk==0.8.0` I run into similar issues with `httpx[http2]`:
```
The conflict is caused by:
    prefect 2.7.0 depends on httpx[http2]>=0.23
    firebolt-sdk 0.8.0 depends on httpx[http2]==0.21.3
```

I'm wondering if it's possible for the upper bound of `pydantic[dotenv]` to be increased (or removed entirely) [here](https://github.com/firebolt-db/firebolt-python-sdk/blob/main/setup.cfg#L31) to allow `pydantic==1.10`, as `httpx[http2]==0.23.0` is no longer an issue (as of `firebolt-sdk==0.13.1`).

stepansergeevitch:
Hello @zzstoatzz,
Thank you for your issue!
I've already relaxed the pydantic version in this PR: https://github.com/firebolt-db/firebolt-python-sdk/pull/232
This will be availble as soon as it will get merged and released

zzstoatzz:
@stepansergeevitch 

great! thank you

stepansergeevitch:
Released in 0.15.0

PR #232, opened 2023-01-23
chore: Relax pydantic version

stepansergeevitch:
Released `pydantic` version to allow `>=1.10` (https://github.com/firebolt-db/firebolt-python-sdk/issues/227)
Added some changes to field definitions in order to support it


ptiurin (approved):
lgtm

PR #233, opened 2023-01-30
feat: Fir 15407 new output format

stepansergeevitch:
Changed output format from `JSONCompact` to `JSON_Compact`. Added support for new Firebolt-style types.


src/firebolt/async_db/_types.py
```diff
@@ -143,88 +143,46 @@ def __eq__(self, other: object) -> bool:
...
+    Date = "date"
+    DateExt = "date_ext"
+    PGDate = "pgdate"
 
-    # DATETIME, TIMESTAMP
-    DateTime = "DateTime"
-    TimestampNtz = "TimestampNtz"
-    TimestampTz = "TimestampTz"
+    Timestamp = "timestamp"
+    TimestampExt = "timestamp_ext"
```
ptiurin, line 164: I think Mosha said `timestamp_ext` is no longer supported. Might as well get rid of it.

ptiurin (approved):
Quick note, otherwise looks good.

src/firebolt/async_db/_types.py
stepansergeevitch: A query like `CAST('2019-07-31 01:01:01.1234' AS TIMESTAMP_EXT(4))` still return `timestamp_ext` for me, I guess we`ll need to keep it for now

PR #235, opened 2023-02-08
feat: Fir 16953 bytea support

stepansergeevitch:
Added support for bytea type (both parsing and formatting)

### DIFF
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -229,7 +229,10 @@ BYTEA_PREFIX = "\\x"
 
 
 def _parse_bytea(str_value: str) -> bytes:
-    if len(str_value) < len(BYTEA_PREFIX) or str_value[:len(BYTEA_PREFIX)] != BYTEA_PREFIX:
+    if (
+        len(str_value) < len(BYTEA_PREFIX)
+        or str_value[: len(BYTEA_PREFIX)] != BYTEA_PREFIX
+    ):
         raise ValueError(f"Invalid bytea value format: {BYTEA_PREFIX} prefix expected")
     return bytes.fromhex(str_value[len(BYTEA_PREFIX) :])
 
--- a/src/firebolt/async_db/_types.py
+++ b/src/firebolt/async_db/_types.py
@@ -49,7 +49,7 @@ from firebolt.utils.exception import (
 from firebolt.utils.util import cached_property
 
 _NoneType = type(None)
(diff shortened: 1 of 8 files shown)


tests/integration/dbapi/async/test_queries_async.py
```diff
@@ -487,3 +493,27 @@ async def test_server_side_async_execution_get_status(
...
+    with connection.cursor() as c:
+        await c.execute("DROP TABLE IF EXISTS test_bytea_roundtrip")
+        await c.execute(
+            "CREATE FACT TABLE test_bytea_roundtrip(id int, b bytea) primary index id"
+        )
+
+        data = "bytea_123"
+
+        await c.execute(
+            "INSERT INTO test_bytea_roundtrip VALUES (1, ?)", (Binary(data),)
```
ptiurin, line 511: I'd add a couple of special characters here as well, just to make sure we handle them correctly. e.g. '\n', '\\', ヽ༼ຈل͜ຈ༽ﾉ

ptiurin (approved):
Otherwise, looks good

tests/integration/dbapi/conftest.py
```diff
@@ -75,7 +75,8 @@ def all_types_query() -> str:
         'true as "boolean", '
         "[1,2,3,4] as \"array\", cast('1231232.123459999990457054844258706536' as "
         'decimal(38,30)) as "decimal", '
-        "cast(null as  int) as nullable"
+        "cast(null as  int) as nullable, "
+        "'abc123'::bytea as bytea"
```
ptiurin, line 79: Just a heads-up, unquoted identifiers are not supposed to work and SQL team has just released a fix. We should change this to `as "bytea"` and any other new types that worked this way.

src/firebolt/async_db/_types.py
```diff
@@ -219,6 +223,13 @@ def parse_type(raw_type: str) -> Union[type, ARRAY, DECIMAL]:  # noqa: C901
         return str
 
 
+BYTEA_PREFIX = "\\x"
+
+
+def _parse_bytea(str_value: str) -> bytes:
+    return bytes.fromhex(str_value[len(BYTEA_PREFIX) :])
```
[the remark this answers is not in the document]
stepansergeevitch: My idea was that it might be a pretty heavy operation on a large scale (since string slice is basically copy). Yet, I think we should be fine in a general case, we have less performant pieces anyway

PR #237, opened 2023-02-15
fix: Decimal class type compatibility

ptiurin:
`__name__` attribute is used in DBT to map python class to SQL query.
This is needed for https://github.com/firebolt-db/dbt-firebolt/pull/95


stepansergeevitch (commented):
Should we add one for `ARRAY` as well?

ptiurin:
> Should we add one for `ARRAY` as well?

Sure, added it. DBT uses slightly different check for array, but might as well keep everything standardised.

stepansergeevitch (approved):

PR #240, opened 2023-03-06
ci: version split

ptiurin:
Splitting CI to run on both main and backwards-compatibility branch `0.x`.

Confirmed most workflows are working:
* Unit tests
* PR verify
* Security-scan
* integration tests

Since other wfs are copies I'm not able to verify they work until they're merged in main, but they use the similar idea to the ones that have been tested.

In checkout action, ref is the branch we want to checkout. If ref is undefined or empty it defaults to the branch the workflow is being called from so`input` parameter to most of our workflows is optional. It's mainly needed for nightlies, which are always run from the main branch.


stepansergeevitch (approved):
Looks very good. Let's have same changes to other driver so we can gradually unblock them

PR #242, opened 2023-03-15
test: Fix insert rowcount

ptiurin:
With the latest changes to aragog API behaviour has changed. Modifying our tests to account for it.

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -296,17 +296,16 @@ class BaseCursor:
     ]:
         """Fetch information about executed query from http response."""
 
-        # Empty response is returned for insert query
-        if response.headers.get("content-length", "") == "0":
-            return (-1, None, None, None)
         try:
             # Skip parsing floats to properly parse them later
             query_data = response.json(parse_float=str)
             rowcount = int(query_data["rows"])
-            descriptions = [
+            descriptions: Optional[List[Column]] = [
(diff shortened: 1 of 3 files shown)


src/firebolt/async_db/cursor.py
```diff
@@ -303,10 +303,12 @@ def _row_set_from_response(
             # Skip parsing floats to properly parse them later
             query_data = response.json(parse_float=str)
             rowcount = int(query_data["rows"])
-            descriptions = [
+            descriptions: Optional[List[Column]] = [
                 Column(d["name"], parse_type(d["type"]), None, None, None, None, None)
                 for d in query_data["meta"]
             ]
+            if not descriptions:
```
stepansergeevitch, line 310: Why can't we just have it empty?

stepansergeevitch: Ok, pep 249 says it should be None

stepansergeevitch (approved):

src/firebolt/async_db/cursor.py
ptiurin: I feel like it's a bit counter-intuitive to receive empty array in description. For example,
A user does an INSERT query, checks cursor.rowcount is correct then checks for some reason `cursor.description` and it's []. The name itself implies singular as well.
I don't feel too strongly about it, but this is also how we had it up until now.

PR #251, opened 2023-05-05
WIP: Refactor sync async

ptiurin:


src/firebolt/async_db/connection.py
```diff
@@ -326,25 +322,34 @@ def __init__(
...
         return c
 
-    async def _aclose(self) -> None:
+    # Context manager support
+    async def __aenter__(self) -> Connection:
+        if self.closed:
+            raise ConnectionClosedError("Connection is already closed.")
+        return self
+
+    def _remove_cursor(self, cursor: Cursor) -> None:
```
stepansergeevitch, line 340: This method, `closed` and `commit` seem to be identical for both classes. Can we move them to the base class

src/firebolt/async_db/connection.py
```diff
@@ -45,6 +47,15 @@
 logger = logging.getLogger(__name__)
 
 
+def _validate_engine_name_and_url(
```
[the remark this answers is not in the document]
ptiurin: The reason I split them was in case we want to actually validate the engine with a request, and since requests would have to be sync/async there's a duplication. However, I think I'm overengineering this, as we're changing this functionality for Identity anyway. I'll consolidate.

src/firebolt/async_db/connection.py
stepansergeevitch: I think it makes sense
The less code duplication we have the better IMO

stepansergeevitch (approved):

PR #253, opened 2023-05-15
test: Fix rowcount tests

ptiurin:
Create/drop queries no longer return a body. Reflecting this in tests.


ptiurin:
Waiting on confirmation that this is expected.

stepansergeevitch (approved):

PR #261, opened 2023-05-30
test: Additional tests to satisfy mutation testing

ptiurin:
Improving test coverage in preparation for mutation test CI.

### DIFF
--- a/tests/unit/async_db/test_connection.py
+++ b/tests/unit/async_db/test_connection.py
@@ -18,6 +18,14 @@ from firebolt.utils.exception import (
 from firebolt.utils.token_storage import TokenSecureStorage
 
 
+async def test_connection_attributes(connection: Connection) -> None:
+    with raises(AttributeError):
+        await connection.not_a_cursor()
+
+    with raises(AttributeError):
+        connection.not_a_database
+
+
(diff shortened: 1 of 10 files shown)


tests/unit/db/conftest.py
```diff
@@ -26,6 +26,24 @@ def connection(
         yield connection
 
 
+@fixture
+def system_connection(
```
stepansergeevitch, line 30: Please rebase on a new master version, there is already this fixture present

tests/unit/async_db/test_connection.py
```diff
@@ -18,6 +18,14 @@
 from firebolt.utils.token_storage import TokenSecureStorage
 
 
+async def test_connection_attributes(connection: Connection) -> None:
```
stepansergeevitch, line 21: Please add a comment on why do we need this test

stepansergeevitch (approved):
Generally LGTM

tests/unit/async_db/test_connection.py
ptiurin: Yeah, I didn't have a chance to properly test this one so it looks odd. Reworked it and marked to skip since __slots__ has no effect in Connection class. :(

PR #265, opened 2023-06-19
docs: Update Connecting_and_queries.rst

pauls-FB:
Add context manager to connection example

### DIFF
--- a/docsrc/Connecting_and_queries.rst
+++ b/docsrc/Connecting_and_queries.rst
@@ -26,6 +26,7 @@ To get started, follow the steps below:
 
 		from firebolt.db import connect
 		from firebolt.client import DEFAULT_API_URL
+		from firebolt.client.auth import UsernamePassword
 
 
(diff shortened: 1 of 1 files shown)


docsrc/Connecting_and_queries.rst
```diff
@@ -66,13 +67,11 @@ To get started, follow the steps below:
...
-					username=username,
-					password=password,
-				)
-
+				with connect(
+    					engine_name=engine_name,
+    					database=database_name,
+    					auth=UsernamePassword(username, password),
+				) as connection:
 				cursor = connection.cursor()
```
ptiurin, line 75: Need to indent this line now, since the previous block starts with `with ...`. Otherwise if someone were to copy the code as is, it will be broken right away.
Same below in dotenv section.

pauls-FB: Added indent

PR #270, opened 2023-07-03
lock pydantic version

stepansergeevitch:
Lock pydantic version to 1.* to avoid incompatibility issues

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,7 +28,7 @@ install_requires =
     appdirs-stubs>=0.1.0
     cryptography>=3.4.0
     httpx[http2]==0.24.0
-    pydantic[dotenv]>=1.8.2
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -28,7 +28,7 @@ install_requires =
     appdirs-stubs>=0.1.0
     cryptography>=3.4.0
     httpx[http2]==0.24.0
-    pydantic[dotenv]>=1.8.2
+    pydantic[dotenv]>=1.8.2,<=1.10.10
```
ptiurin, line 31: Should we do `<2.0` instead, so if there are any patches for 1.x we'll support them?

stepansergeevitch: makes sense

PR #276, opened 2023-07-05
test: Remove deprecated types

ptiurin:


stepansergeevitch (approved):
Please also remove deprecated types from _InternalTypes enum
https://github.com/firebolt-db/firebolt-python-sdk/blob/0.x/src/firebolt/common/_types.py#L167
https://github.com/firebolt-db/firebolt-python-sdk/blob/0.x/src/firebolt/common/_types.py#L171

PR #277, opened 2023-07-05
feat: Performance debug info

RotemFB:
added a timer option that counts the time of Authentication, Resolving engine name, Running query and Parsing query output. the info will be written on the log if an env variable named `FIREBOLT_SDK_PERFORMANCE_DEBUG` will be equal to 1

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import logging
 import socket
+from os import environ
+from time import time
 from types import TracebackType
 from typing import Any, Dict, List, Optional
 
@@ -231,9 +233,21 @@ async def connect(
 
     else:
         try:
+            timerStart = time()
             engine_url, status, attached_db = await _get_engine_url_status_db(
                 system_engine_connection, engine_name
             )
+
+            teimerEnd = time()
+            envVar = "0"
+            try:
+                envVar = environ["FIREBOLT_SDK_PERFORMANCE_DEBUG"]
+            except Exception:
+                pass
+
+            if envVar == "1":
+                totalTime = "{:.2f}".format(round((teimerEnd - timerStart), 2))
+                logger.debug(f"[PERFORMANCE] Resolving engine name {totalTime}s")
 
             if status != "Running":
                 raise EngineNotRunningError(engine_name)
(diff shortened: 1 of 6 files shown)


src/firebolt/async_db/connection.py
```diff
@@ -231,10 +233,22 @@ async def connect(
 
     else:
         try:
+            timerStart = time()
             engine_url, status, attached_db = await _get_engine_url_status_db(
                 system_engine_connection, engine_name
             )
 
+            teimerEnd = time()
```
stepansergeevitch, line 241: let's create a separate function with this logic and use it everywhere
I would suggest to consider implementing a context manager for this purpose
https://peps.python.org/pep-0343/

src/firebolt/async_db/connection.py
```diff
@@ -231,10 +233,22 @@ async def connect(
...
+            envVar = "0"
+            try:
+                envVar = environ["FIREBOLT_SDK_PERFORMANCE_DEBUG"]
+            except Exception:
+                pass
```
stepansergeevitch, line 246: Can we replace this logic with `environ.get("FIREBOLT_SDK_PERFORMANCE_DEBUG", "0")`?

src/firebolt/async_db/cursor.py
```diff
@@ -199,14 +201,23 @@ async def _do_execute(
...
+                                "async_execution": 1,
+                                "advanced_mode": 1,
+                                "output_format": JSON_OUTPUT_FORMAT,
+                            },
+                        )
+
+                    envVar = environ.get("FIREBOLT_SDK_PERFORMANCE_DEBUG", "0")
+                    if envVar == "1":
+                        logger.debug(
+                            f"[PERFORMANCE] Running query {query[:50]} ... {runTime}s"
```
stepansergeevitch, line 218: Do not add dots `...` in case the query length is less then 50 characters. It will look weird then

src/firebolt/common/base_cursor.py
```diff
@@ -367,7 +368,11 @@ def fetchall(self) -> List[List[ColType]]:
         left, right = self._get_next_range(self.rowcount)
         assert self._rows is not None
         rows = self._rows[left:right]
-        return [self._parse_row(row) for row in rows]
+
+        with Timer("[PERFORMANCE] Parsing query output into native Python types "):
+            result = [self._parse_row(row) for row in rows]
+
+        return result
```
ptiurin, line 375: We would want this in `fetchone` and `fetchmany` functions as well since users could be fetching results in chunks.

ptiurin (approved):
lgtm

PR #279, opened 2023-07-10
feat: show error body

RotemFB:
added an error body when there is an unfamiliar error that has some info in it

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -40,6 +40,7 @@ from firebolt.utils.exception import (
     CursorClosedError,
     EngineNotRunningError,
     FireboltDatabaseError,
+    FireboltError,
     OperationalError,
     ProgrammingError,
     QueryNotRunError,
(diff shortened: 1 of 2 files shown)


src/firebolt/async_db/cursor.py
```diff
@@ -133,6 +133,12 @@ async def _raise_if_error(self, resp: Response) -> None:
                     f"Firebolt engine {self.connection.engine_url} "
                     "needs to be running to run queries against it."
                 )
+        if (
+            codes.is_error(resp.status_code)
+            and "Content-Length" in resp.headers
+            and int(resp.headers["Content-Length"]) > 0
+        ):
+            logger.debug(f"Something went wrong: {resp.read().decode('utf-8')}")
```
ptiurin, line 141: I would surround this with try/catch. While Content-Length is usually a good indicator the body is there when everything goes well it's not always the case when an error is thrown. I'd rather fall back to `raise_for_status` in the next line when something goes wrong than the user getting `can't decode character at line 0`.

stepansergeevitch (approved):
LGTM

PR #280, opened 2023-07-11
fix: compatibility with httpcore 0.17.3

RotemFB:
fixed the imports from httpcore so it will be compatible for 0.17.3 version


ptiurin:
Are we backwards-compatible? i.e. Does this still work with `httpcore<0.17.3`?

ptiurin (approved):
As discussed over the daily, LGTM
If pinning a minimum version causes issues we can always work around it.

PR #281, opened 2023-07-13
ci: run only failed tests when re-running

RotemFB:
using cache in the ci in order to re-run only the failed tests in order to save some time


ptiurin (approved):
lgtm, just trim the whitespace in `unit-test.yml` to keep the changelog clean

PR #283, opened 2023-07-17
ci: use new identity setup action for integration tests

stepansergeevitch:
Updated integration-testing-setup action, that's used for db and engine setup, to v2


.github/workflows/nightly-workflow.yml
```diff
@@ -55,7 +55,7 @@ jobs:
 
       - name: Setup database and engine
         id: setup
-        uses: firebolt-db/integration-testing-setup@v1
+        uses: firebolt-db/integration-testing-setup@v2
         with:
           firebolt-username: ${{ secrets.FIREBOLT_STG_USERNAME }}
           firebolt-password: ${{ secrets.FIREBOLT_STG_PASSWORD }}
```
ptiurin, line 61: We probably need new inputs here as well. e.g. `firebolt-client-id:` etc.

ptiurin (approved):
Apart from that lgtm.

PR #295, opened 2023-08-10
fix: delete database when engines starting or stopping

RotemFB:
when deleting a database the check if the attached engines are starting or stopping was wrong

### DIFF
--- a/src/firebolt/model/database.py
+++ b/src/firebolt/model/database.py
@@ -88,7 +88,7 @@ class Database(FireboltBaseModel):
         """
 
         for engine in self.get_attached_engines():
-            if engine.current_status not in {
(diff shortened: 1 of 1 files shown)


ptiurin:
Would you be able to add some tests for this?

stepansergeevitch (approved):

tests/unit/service/test_database.py
ptiurin, line 1: I was thinking more of checking for `AttachedEngineInUseError`, which was fixed by this PR. e.g. Check that if `get_attached_engines` returns starting/stopping status on delete the error is raised, otherwise the call succeeds.

ptiurin (approved):
lgtm

PR #310, opened 2023-09-01
feat: Add rm support for engine type update

stepansergeevitch:
Added `engine_type` parameter to `engine.update` to allow changing engine type via resource management api

### DIFF
--- a/tests/integration/resource_manager/test_engine.py
+++ b/tests/integration/resource_manager/test_engine.py
@@ -2,7 +2,7 @@ from collections import namedtuple
 
 from firebolt.client.auth import Auth
 from firebolt.service.manager import ResourceManager
-from firebolt.service.types import EngineStatus
+from firebolt.service.types import EngineStatus, EngineType, WarmupMethod
 
 
 def make_engine_name(database_name: str, suffix: str) -> str:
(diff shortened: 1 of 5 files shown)


tests/integration/resource_manager/test_engine.py
```diff
@@ -42,10 +42,11 @@ def test_create_start_stop_engine(
 ParamValue = namedtuple("ParamValue", "set expected")
 ENGINE_UPDATE_PARAMS = {
     # commented parameters are not available yet
```
ptiurin, line 44: nit: We can remove this comment now that parameters are available to avoid confusion.

tests/integration/resource_manager/test_engine.py
```diff
@@ -42,10 +42,11 @@ def test_create_start_stop_engine(
...
-    #    "scale": ParamValue(23, 23),
-    #    "spec": ParamValue("B1", "B1"),
+    "scale": ParamValue(23, 23),
```
ptiurin, line 45: We're testing on x23? Should we use something $$ safer, in case it's started for some reason?

ptiurin (approved):
Some minor clarifications, otherwise LGTM

tests/integration/resource_manager/test_engine.py
stepansergeevitch: It's never started so we should be safe

stepansergeevitch: I can reduce the size though anyway

PR #319, opened 2023-10-19
fix: Remove redundant lock

ptiurin:
Similar to the change in 1.x, removing trino lock as it affects runs from within asyncio.

Cherry picked from https://github.com/firebolt-db/firebolt-python-sdk/pull/303

UPD: this does not fix the asyncio runs. Added a test that shows this.


ptiurin:
Added a failing test that shows we can't run via asyncio.run

stepansergeevitch (approved):

ptiurin (commented):
nice one, lgtm

PR #322, opened 2023-11-03
docs: pip install upper version limit

stepansergeevitch:
Added an upper limit for 0.* python sdk version

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 ### Installation
 
 * Requires Python `>=3.7`
-* `pip install firebolt-sdk`
+* `pip install "firebolt-sdk<1.0.0a1"`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@
 ### Installation
 
 * Requires Python `>=3.7`
-* `pip install firebolt-sdk`
```
ptiurin, line 13: Should we do `"firebolt-sdk==0.*"` instead? It's just I feel uneasy relying on an alpha always being there.

stepansergeevitch: It makes sense

stepansergeevitch: Done

ptiurin (approved):

PR #323, opened 2023-11-19
feat: backwards-compatible version merge

ptiurin:
FIR-28091
[Design doc](https://docs.google.com/document/d/1jmvD7x1jYETtE0KL4YcTdy1AZX5S5a8YU-2ybpByw7c/edit#heading=h.gfozuvdf3bwx)

Merging versions compatible with Firebolt 1.0 and Firebolt 2.0, essentially making the driver backwards-compatible.

Some decisions made:
* Version names kept as V1 and V2 to be consistent across drivers and correspond to Firebolt versions.
* Tests are grouped in V1 and V2 directories with shared sections outside of those dirs.
* Tagging was considered, but grouping tests in respective dirs felt so much cleaner and the ease of selecting then in pytest with `-k "not V1"` or `-k "not V2"` is pretty much the same.
* There might be some duplication in tests, but it can be refactored at a later date. The most important part is that code is covered.
* Resource manager only has engine get/start/stop for V1 and V2 kept as is. This is to be compatible with airflow. The only incompatible client is firebolt-cli, which uses a wide range of RM features for V1.
* Many empty __init__.py files were added to aid Python module discovery when file names (or test names) are the same.
* Engine start/stop integration test is skipped as it will interfere with the other tests.
* `test_from_asyncio` leads to an error so I've removed `async_cached_property` and implemented adhoc caching. Worst case scenario it will lead to duplicate round-trip to the server so no locking is strictly necessary.


ptiurin:
Integration tests successful: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/6947833011

stepansergeevitch (approved):

PR #328, opened 2023-11-29
feat: support USE DATABASE query

ptiurin:
Enabling USE DATABASE recognition in Firebolt SDK.

This PR sets the relevant protocol version in a base class of Client, since this functionality is shared across V1/V2 sync and async implementations.
Also, to facilitate this change of behaviour we no longer read database attribute from the connection, but store it on the cursor itself. Therefore if one cursor changes the database it does not affect other cursors from the same connection.
This approach is extensible to potentially include engine, once the design is finalised as we check allowed parameters that are returned from the server and add those to the query params dict.

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -82,6 +82,12 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
         super().__init__(*args, **kwargs)
         self._client = client
         self.connection = connection
+        if connection.database:
+            self.parameters["database"] = connection.database
+
+    @property
+    def database(self) -> Optional[str]:
+        return self.parameters.get("database")
 
     @abstractmethod
     async def _api_request(
@@ -100,11 +106,11 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
                 f"Error executing query:\n{resp.read().decode('utf-8')}"
             )
         if resp.status_code == codes.FORBIDDEN:
-            if self.connection.database and not await self.is_db_available(
-                self.connection.database
+            if self.parameters["database"] and not await self.is_db_available(
+                self.parameters["database"]
             ):
                 raise FireboltDatabaseError(
-                    f"Database {self.connection.database} does not exist"
+                    f"Database {self.parameters['database']} does not exist"
                 )
             raise ProgrammingError(resp.read().decode("utf-8"))
         if (
@@ -201,6 +207,8 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
                         query, {"output_format": JSON_OUTPUT_FORMAT}
                     )
                     await self._raise_if_error(resp)
+                    # get parameters from response
+                    self._parse_response_headers(resp.headers)
                     row_set = self._row_set_from_response(resp)
 
                 self._append_row_set(row_set)
@@ -440,8 +448,8 @@ class CursorV2(Cursor):
         parameters = parameters or {}
         if use_set_parameters:
             parameters = {**(self._set_parameters or {}), **parameters}
-        if self.connection.database:
-            parameters["database"] = self.connection.database
+        if self.parameters:
+            parameters = {**self.parameters, **parameters}
         if self.connection._is_system:
             assert isinstance(self._client, AsyncClientV2)
             parameters["account_id"] = await self._client.account_id
@@ -550,7 +558,7 @@ class CursorV1(Cursor):
             url=f"/{path}",
             method="POST",
             params={
-                "database": self.connection.database,
+                "database": self.parameters["database"],
                 **(parameters or dict()),
             },
             content=query,
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -188,7 +188,6 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
                             query,
                             {
                                 "async_execution": 1,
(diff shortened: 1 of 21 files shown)


ptiurin:
Reminder to myself: don't delete the branch until after the demo of USE database functionality.

tests/integration/dbapi/async/V2/test_queries_async.py
```diff
@@ -417,3 +419,41 @@ async def test_bytea_roundtrip(
...
+async def setup_db(connection_system_engine_no_db: Connection, use_db_name: str):
+    use_db_name = use_db_name + "_async"
+    with connection_system_engine_no_db.cursor() as cursor:
+        await cursor.execute(f"CREATE DATABASE {use_db_name}")
+        yield
+        await cursor.execute(f"DROP DATABASE {use_db_name}")
+
+
+@mark.xfail("dev" not in environ[API_ENDPOINT_ENV], reason="Only works on dev")
+async def test_use_database(
```
alexradzin, line 434: Is it possible to reuse the code between v1 and v2 even for tests? The functions `test_use_database()` look very similar.

tests/unit/async_db/V2/test_cursor.py
```diff
@@ -841,3 +841,24 @@ async def test_cursor_iterate(
     with raises(CursorClosedError):
         async for res in cursor:
             pass
+
+
+async def test_server_side_header_database(
```
alexradzin, line 846: Code reuse?

alexradzin (approved):
approved with some comments

src/firebolt/async_db/cursor.py
```diff
@@ -100,11 +106,11 @@ async def _raise_if_error(self, resp: Response) -> None:
                 f"Error executing query:\n{resp.read().decode('utf-8')}"
             )
         if resp.status_code == codes.FORBIDDEN:
-            if self.connection.database and not await self.is_db_available(
-                self.connection.database
+            if self.parameters["database"] and not await self.is_db_available(
```
stepansergeevitch, line 109: I think we should better use `self.database` here. Also, I believe that if "database" is not in parameter it will fail with an exception.

stepansergeevitch: Here and in other places

src/firebolt/async_db/cursor.py
```diff
@@ -550,7 +558,7 @@ async def _api_request(
             url=f"/{path}",
             method="POST",
             params={
-                "database": self.connection.database,
+                "database": self.parameters["database"],
```
stepansergeevitch, line 561: Do we still need to add database separately? Isn't it a part of self.parameters?

src/firebolt/common/base_cursor.py
```diff
@@ -243,6 +248,17 @@ def _reset(self) -> None:
         self._next_set_idx = 0
         self._query_id = ""
 
+    def _parse_response_headers(self, headers: Headers) -> None:
+        """Parse response and update relevant cursor fields."""
+        update_parameters = headers.get("Firebolt-Update-Parameters")
+        # parse update parameters dict and set keys as attributes
+        if update_parameters:
+            # parse key1=value1,key2=value2 comma separated string into dict
+            param_dict = dict(item.split("=") for item in update_parameters.split(","))
```
stepansergeevitch, line 257: Just a suggestion, maybe we can log with a debug level if received parameter is not known. Might help with debugging in some cases

src/firebolt/async_db/cursor.py
ptiurin: You're right. Somehow this _api_request was different from the others.

src/firebolt/common/base_cursor.py
ptiurin: Fair point.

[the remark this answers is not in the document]
ptiurin: Fair, might as well do it. Can't trust the server side not to mess this up.

src/firebolt/client/client.py
```diff
@@ -51,6 +51,11 @@ def __init__(
         self._api_endpoint = URL(fix_url_schema(api_endpoint))
         self._auth_endpoint = get_auth_endpoint(self._api_endpoint)
         super().__init__(*args, auth=auth, **kwargs)
+        self._set_default_header("Firebolt-Protocol-Version", PROTOCOL_VERSION)
```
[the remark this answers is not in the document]
ptiurin: Agreed, having it as a constant is better.

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ptiurin: Switched to an attribute instead.

tests/integration/dbapi/async/V2/test_queries_async.py
ptiurin: While it is possible, some of the fixtures those tests use are V1/V2 specific. If we move this test up above the v1/v2 directories and call the shared code from inside it will create an indirection that's really hard to follow, since it's evaluated at runtime. I'd rather have duplication in test code than obfuscation.
image

tests/unit/async_db/V2/test_cursor.py
ptiurin: Same as above. It would also be easier to remove V1 afterwards.

stepansergeevitch (approved):

PR #332, opened 2023-12-21
build: Relax pydantic version requirement

ptiurin:
Part of refactoring needed for MWAA(managed airflow) support in FIR-28712
This way we'll be able to support both legacy and the latest managed Airflow deployment.

Ticket FIR-29465

Note: there's a difference between pydantic validation and mypy validation of pydantic models. The former respects `populate_by_name` and `allow_population_by_field_name` and the latter does not understand it. Installing pydantic.mypy plugin reverses this, it assumes either alias is correct or the kay. There's some discussion in mypy about this but basically nothing will change until v3.
Currently, it's safe to assume pydantic will populate data from the server correctly even if it comes with aliases, but for the code initialisation we're sticking to keys.


stepansergeevitch (approved):

ptiurin:
> Is there any work going on for this PR? I'm using `firebolt-sdk` in my project and would be nice to be able to use Pydantic v2.

Hi. Yes, just need to update it to reflect new changes in the main. Should be able to get to it in a couple of days.

vgutkovsk:
> > Is there any work going on for this PR? I'm using `firebolt-sdk` in my project and would be nice to be able to use Pydantic v2.
> 
> Hi. Yes, just need to update it to reflect new changes in the main. Should be able to get to it in a couple of days.

Thanks! Looking forward to release.

stepansergeevitch (approved):
LGTM

ptiurin:
Released in [1.2.2](https://github.com/firebolt-db/firebolt-python-sdk/releases/tag/1.2.2)

PR #333, opened 2023-12-27
fix: Readd database support for resource manager

ptiurin:


src/firebolt/service/V1/region.py
```diff
@@ -0,0 +1,65 @@
...
+
+        return {r.key: r for r in self.regions}
+
+    @cached_property
+    def default_region(self) -> Region:
+        """Default AWS region, could be provided from environment."""
+
+        if not self.default_region_setting:
+            raise ValueError(
+                "The environment variable FIREBOLT_DEFAULT_REGION must be set."
```
stepansergeevitch, line 46: Should we update this with `default_region variable should be provided`? I don't think we even support that env variable anymore

stepansergeevitch (approved):
LGTM, left a single comment

PR #335, opened 2023-12-28
docs: Update to reflect the current state of the code

ptiurin:
Added some of the missing sections on async.
Updated the auth to suggest Settings is deprecated.
Removed some of the missing engine functions.
We had a mix of tabs and spaces so I've converted everything to spaces.

### DIFF
--- a/docsrc/Managing_resources.rst
+++ b/docsrc/Managing_resources.rst
@@ -23,9 +23,50 @@ To get started, follow the steps below:
 
 	::
 
+        from firebolt.client.auth import UsernamePassword
 		from firebolt.service.manager import ResourceManager
-		from firebolt.common import Settings
 
+**2. Initialize a ResourceManager object**
+
+    A ResourceManager object contains the user credentials and other information needed to
+    manage Firebolt databases and engines.
+
+    ResourceManager uses the following parameters:
+
+    +---------------------+-----------------------------------------------------------------------------------------------------------------------------+
+    | ``auth``            |  Auth object, containing your credentials. See :ref:`Auth <firebolt.client:auth>` for more details.                         |
+    +---------------------+-----------------------------------------------------------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


ptiurin:
Built doc: https://python-sdk.docs.firebolt.io/en/fix-doc-render/

docsrc/Managing_resources.rst
```diff
@@ -17,80 +17,118 @@ To get started, follow the steps below:
...
+.. _required_resourcemanager_imports:
+
+    ::
+
+        from firebolt.client.auth import UsernamePassword
+        from firebolt.service.manager import ResourceManager
 
 **2. Initialize a Settings object**
 
-	A Settings object contains the user credentials and other information needed to
```
stepansergeevitch, line 32: I belive we shouldn't mention `Settings` for a 1.x version since it's deprecated and shouldn't be used

docsrc/Managing_resources.rst
```diff
@@ -17,80 +17,118 @@ To get started, follow the steps below:
...
+            ``.env`` file if it is set to ``None`` instead of having values, as shown below:
 
-			::
+            ::
 
-				settings = None
+                settings = None
 
 
 **3. Initialize a ResourceManager object**
```
stepansergeevitch, line 118: I believe this section should also be removed since we already have one and this one contains outdated information

docsrc/Managing_resources.rst
ptiurin: This is a PR for 0.x branch, 1.x does not have any mention of settings - https://github.com/firebolt-db/firebolt-python-sdk/blob/main/docsrc/Managing_resources.rst
This RM setup section comes second after the one with direct parameters and is labeled `(legacy)`. Some users of 0.x might still be using Settings so we should keep this until FB1.0 is deprecated altogether.

docsrc/Managing_resources.rst
ptiurin: This is a separate subsection of a larger `Setting up a ResourceManager object (legacy)` section, which describes a legacy way to connect.

PR #342, opened 2024-01-24
fix: Fir 27986 engine parameters formatting

stepansergeevitch:
Fixed parameters formatting for FB2.0 engines service

### DIFF
--- a/src/firebolt/service/V2/engine.py
+++ b/src/firebolt/service/V2/engine.py
@@ -119,12 +119,12 @@ class EngineService(BaseService):
     def _format_engine_parameter(
         value: Union[str, int, EngineType, InstanceType, WarmupMethod]
     ) -> Union[str, int]:
+        if not isinstance(value, (str, int, EngineType, InstanceType, WarmupMethod)):
+            raise TypeError(f"Unsupported type {type(value)} for engine parameter. ")
         if isinstance(value, InstanceType):
             return value.name
         if isinstance(value, (EngineType, WarmupMethod)):
             return str(value.value)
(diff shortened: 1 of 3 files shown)


src/firebolt/service/V2/engine.py
```diff
@@ -115,6 +115,18 @@ def get_many(
             ]
             return [Engine._from_dict(_dict, self) for _dict in dicts]
 
+    @staticmethod
+    def _format_engine_parameter(
+        value: Union[str, int, EngineType, InstanceType, WarmupMethod]
```
ptiurin, line 120: nit: we should include `float` here

src/firebolt/service/V2/engine.py
```diff
@@ -115,6 +115,18 @@ def get_many(
...
+    ) -> Union[str, int]:
+        if isinstance(value, InstanceType):
+            return value.name
+        if isinstance(value, (EngineType, WarmupMethod)):
+            return str(value.value)
+        elif isinstance(value, (int, float)):
+            return int(value)
```
ptiurin, line 127: I'm not sure about the float here. If someone passes `1.1` in scale, that's a misconfiguration and should explicitly fail. I think the only way this is ok if a float value has no decimal part e.g. `1.0`

src/firebolt/service/V2/engine.py
stepansergeevitch: I would better not allow float. This would remove any confusion, there is no need to use it anyways

ptiurin (approved):

PR #344, opened 2024-01-30
ci: remove env parameter from ci

stepansergeevitch:
Removed `environment` parameter from all CI jobs, and set `staging` as a default env


stepansergeevitch:
Integration tests: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/7711319572/job/21016452424

ptiurin (approved):

PR #346, opened 2024-01-31
fix: resolve cyclic dependency

stepansergeevitch:
Fixed cyclic import error when importing `FireboltError` from `firebolt.common.exceptions`


ptiurin (approved):
lgtm

PR #348, opened 2024-02-07
feat: Disallow internal set parameters

ptiurin:
FIR-30143
Some set parameters are not allowed. Adding an appropriate error message for them.


ptiurin:
I feel there are no appropriate integration tests here since this is purely client-side change. Integration tests would just test the same thing and look exactly the same as the unit tests.

stepansergeevitch (approved):
LGTM, left one comment

PR #352, opened 2024-02-09
feat: Receiving account version 2

ptiurin:
Part of FIR-30151

Adding support for a new field in account resolution call. It will be used in the future to adjust error checking and internal logic.

Setting it to "private" to indicate this should not be directly touched by the user.

Firebolt V1 uses a different url to resolve account name so I'm hardcoding this parameter for completion.


ptiurin:
TODO: add org variables for account V1 and V2

tests/integration/dbapi/async/V2/test_system_engine_async.py
```diff
@@ -16,6 +16,7 @@ async def test_system_engine(
     timezone_name: str,
 ) -> None:
     """Connecting with engine name is handled properly."""
+    connection_system_engine._client._account_version = 1, "Invalid account version"
```
stepansergeevitch, line 19: should it be assert?

ptiurin: whoops, you're right

PR #354, opened 2024-02-15
feat: Use engine support

ptiurin:
FIR-30151

Support decoupling of databases and engines (USE DATABASE, USE ENGINES).

Design decisions:
* Connection object holds initial database/engine that's specified by the user. All cursors created from it inherit these, but also keep them local. This means that cursors change on execution of "USE DATABASE/ENGINE" where they're connected to, but any new cursors from a connection will start from the initial configuration.
* `init_parameters` in connection used for parameters that server side returns from USE command. On connect() we execute these queries on system engine if the user supplied database/engine and pass the resulting parameters to the Connection constructor. This ensures connection is initialised as per user parameters.
* ENGINE_STATUS is now a list, since engines V2 and engines V1 statuses are different.
* Common properties (database) in cursor moved to the base class, since they're used in both cursor versions.
* Response header parsing is refactored to a base class since it's common for sync and async.

### DIFF
--- a/src/firebolt/async_db/util.py
+++ b/src/firebolt/async_db/util.py
@@ -11,8 +11,6 @@ from firebolt.utils.exception import (
 )
 from firebolt.utils.urls import GATEWAY_HOST_BY_ACCOUNT_NAME
 
-ENGINE_STATUS_RUNNING_LIST = ["Running", "ENGINE_STATE_RUNNING"]
-
 
 async def _get_system_engine_url(
     auth: Auth,
--- a/src/firebolt/async_db/util.py
+++ b/src/firebolt/async_db/util.py
@@ -11,7 +11,7 @@ from firebolt.utils.exception import (
 )
 from firebolt.utils.urls import GATEWAY_HOST_BY_ACCOUNT_NAME
 
-ENGINE_STATUS_RUNNING = "Running"
+ENGINE_STATUS_RUNNING_LIST = ["Running", "ENGINE_STATE_RUNNING"]
 
 
 async def _get_system_engine_url(
--- a/src/firebolt/common/base_cursor.py
+++ b/src/firebolt/common/base_cursor.py
@@ -24,6 +24,7 @@ from firebolt.utils.exception import (
     ConfigurationError,
     CursorClosedError,
     DataError,
+    OperationalError,
     QueryNotRunError,
 )
 from firebolt.utils.util import Timer, fix_url_schema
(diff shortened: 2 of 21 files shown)


src/firebolt/common/base_cursor.py
```diff
@@ -273,25 +305,37 @@ def _reset(self) -> None:
...
+                )
+
+    def get_engine_name(self, engine_url: str) -> str:
+        """
+        Get the name of the engine that we're using.
+
+        Args:
+            engine_url (str): URL of the engine
+        """
+        if self._set_parameters.get("engine"):
```
stepansergeevitch, line 336: Can we have `engine` in _set_parameters?

src/firebolt/async_db/util.py
```diff
@@ -11,7 +11,7 @@
 )
 from firebolt.utils.urls import GATEWAY_HOST_BY_ACCOUNT_NAME
 
-ENGINE_STATUS_RUNNING = "Running"
+ENGINE_STATUS_RUNNING_LIST = ["Running", "ENGINE_STATE_RUNNING"]
```
stepansergeevitch, line 14: nit: Move to `common` to avoid duplication

src/firebolt/common/base_cursor.py
```diff
@@ -273,25 +305,37 @@ def _reset(self) -> None:
...
+                    "It will be ignored."
+                )
+
+    def _update_server_parameters(self, parameters: Dict[str, Any]) -> None:
+        for key, value in parameters.items():
+            if key in SERVER_SIDE_PARAMETERS:
+                self.parameters[key] = value
+            else:
+                logger.debug(
+                    f"Trying to set an unknown parameter {key}. " "It will be ignored."
```
stepansergeevitch, line 326: nit: Convert two strings into one

src/firebolt/common/base_cursor.py
stepansergeevitch: In case we can't, convert this function into a property

ptiurin: Yes, _set_parameters populated from a "Firebolt-Update-Endpoint" header returned when `USE ENGINE` is called, returning the gateway URL and engine=X parameter which specifies which engine we're using.

stepansergeevitch: Oh, ok. But in this case am I right that Firebolt-Reset-Session header would also remove the `engine` parameter from _set_statements?

ptiurin: That was exactly my q in https://docs.google.com/document/d/1f0kdPCPbhCrmMTK-sfqAkqEkonT7xyNK-teJmXE_6p0/edit?disco=AAABFIz4oXU
Do you think we should make "engine" a special case?

ptiurin: Ok, thinking about it it really should be "special". Reworked this part a little.

src/firebolt/common/base_cursor.py
```diff
@@ -273,25 +304,38 @@ def _reset(self) -> None:
...
+        }
+        user_parameters = {
+            key: value
+            for key, value in parameters.items()
+            if key not in IMMUATABLE_PARAMETER_LIST
+        }
+
+        self._update_server_parameters(immutable_parameters)
+
+        for key, value in user_parameters.items():
```
stepansergeevitch, line 322: nit: just do `self._set_parameters.update(user_parameters)`?

stepansergeevitch (approved):

src/firebolt/common/base_cursor.py
ptiurin: Reworked again, as discussed offline. switched _set_parameters to parameters and added a test for this.

src/firebolt/common/base_cursor.py
ptiurin: You're right, forgot Python has this syntax.

PR #355, opened 2024-02-16
feat: Add compatible methods for Pydantic V1.8 and V2

lucianosrp:
# Goal

The goal of this PR is to maintain backward compatibility with Pydanitc V1 and V2 without raising any deprecation warnings.

# Description 

This PR allows to use both V1.8 and V2 without triggering any deprecation warnings
It also allows to keep the current code for V3 (*possibly*)

The main edits are done in [src/firebolt/model/V1/__init__.py](https://github.com/firebolt-db/firebolt-python-sdk/compare/main...lucianosrp:firebolt-python-sdk:feat-pydantic-compatibility?expand=1#diff-782ba68728568d92ab7859f77ca1133c790f42abc046f8a379574fceb6bb9f7a) 

It introduces a utility function that gets the desired method based on the current pydantic version:
```python 

GenericCallable: TypeAlias = Callable[..., Any]

# Using `.VERSION` instead of `.__version__` for backward compatibility:
PYDANTIC_VERSION: Final[int] = int(pydantic.VERSION[0])


def use_if_version_ge(
    version_ge: int,
    obj: Union[pydantic.BaseModel, Type[Model]],
    previous_method: str,
    latest_method: str,
) -> GenericCallable:
    
    if PYDANTIC_VERSION >= version_ge:
        return getattr(obj, latest_method)
    else:
        return getattr(obj, previous_method)
```

BaseModel configurations are also handled differently:
```python
class FireboltBaseModel(pydantic.BaseModel):
    if PYDANTIC_VERSION >= 2:
        # Pydantic V2 config
        model_config = ConfigDict(populate_by_name=True, from_attributes=True)

    else:
        # Using Pydantic 1.* config class for backwards compatibility
        class Config:
            extra = "forbid"
            allow_population_by_field_name = True  # Pydantic 1.8
```

# Other changes

 - All calls to `dict` are changed to the new version-compatible method `model_dict`
 - All calls to `parse_obj` are changed to the new version-compatible method `parse_model`
 - All calls to `json` are changed to the new version-compatible method `model_dump_json`

# Tests

Tested with `pydantc==1.8.2` and `pydantic==2.6` on Python 3.9


lucianosrp:
@ptiurin

ptiurin:
Thank you for the contribution @lucianosrp !

Couple of things I've noticed:
- looks like some of the files were converted to the Windows line endings (CRLF), could you revert it back to LF so the diffs would show the actual code differences?
- Unit tests are failing, please have a look at running `pytest tests/unit` 
- Mypy is failing for what I believe is the same issue as the tests. Have a look at the output of `mypy src` from the root of the repo.

lucianosrp:
Great, thanks!

I noticed it is due to the checks being run on python 3.8 instead of 3.9
On my local environmet with `3.9` mypy and pytest are running fine.

I can still procceed to change that line! 

I will revert back to LF !

src/firebolt/service/V1/provider.py
ptiurin, line 1: I believe this file also has line endings changed. Let's change them back.

lucianosrp: You are right - sorry again! 
Fixing it right away

ptiurin (approved):

PR #359, opened 2024-03-01
fix: Async auth lock

ptiurin:
FIR-30447

More intelligent auth lock that doesn't lock on each request, but only when auth flow is executed.
Added tests that verify that we're truly async.


ptiurin:
I want to add tests before marking it ready for review again.

stepansergeevitch (approved):

PR #362, opened 2024-03-11
chore: Update alter database syntax

ptiurin:
FIR-30871

`ALTER DATABASE x WITH` has been changed to `ALTER DATABASE x SET`
Reflecting this in code and adding missing tests.

### DIFF
--- a/tests/unit/service/V2/test_database.py
+++ b/tests/unit/service/V2/test_database.py
@@ -1,5 +1,6 @@
-from typing import Callable
+from typing import Any, Callable, Dict
 
+from httpx import Response, codes
 from pytest import raises
 from pytest_httpx import HTTPXMock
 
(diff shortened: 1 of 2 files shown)


tests/unit/service/V2/test_database.py
```diff
@@ -128,3 +129,50 @@ def test_database_delete_busy_engine(
...
+    httpx_mock.add_callback(
+        update_query_callback, url=system_engine_no_db_query_url, method="POST"
+    )
+    # mock get_attached_engines method to return an engine in a running state
+    mock_database.get_attached_engines = lambda: [mock_engine]
+
+    mocked_service = resource_manager.engines
+    mock_database._service = mocked_service
+
+    # Call the update method
```
alexradzin, line 162: Does this (and similar) comment(s) help to understand the code? IMHO line
```
updated_database = mock_database.update(description="new description")
```
updates description of mock DB and this is obvious even for person that has written 2 python lines in his like (like me) ;)

alexradzin (approved):
Approving, although I am the last person in the world that should approve Python code. 
Also, as far as I understand most of the changes here are in tests that were actually missing. IMHO add this fact to the description of the commit, e.g. "changed keyword WITH to SET in ALTER DATABASE command and added missing tests"

tests/unit/service/V2/test_database.py
ptiurin: You're right. Had those left over from the design phase.

ptiurin:
> Also, as far as I understand most of the changes here are in tests that were actually missing. IMHO add this fact to the description of the commit, e.g. "changed keyword WITH to SET in ALTER DATABASE command and added missing tests"

Since adding/fixing tests is an expectation for every change that's made, I don't think it makes sense to mention it each time in the commit. Commit name should capture the intent with which the changes were made. The diff, on the other hand, describes what was changed in the code.

ptiurin:
> Approving, although I am the last person in the world that should approve Python code.

I disagree. Even if Python is not your primary language you can still spot issues with readability, test coverage or ask questions about logic that's changed.

PR #363, opened 2024-03-18
test: Fix cleanup on v2

ptiurin:
Some dbs were not cleaned up.

Reusing existing fixtures that cleanup the dbs automatically.


ptiurin:
Confirmed no dbs left after integration tests run. Merging.

PR #366, opened 2024-03-28
fix: Fir 31676 python sdk resource manager errors when referencing engines

stepansergeevitch:
Added support for new uppercase engine statuses


src/firebolt/service/V2/types.py
```diff
@@ -53,6 +53,10 @@ class EngineStatus(Enum):
     def __str__(self) -> str:
         return self.value
 
+    @classmethod
+    def from_string(cls, value: str) -> "EngineStatus":
+        return cls(value.capitalize())
```
ptiurin, line 58: I'm just worried that now we have two ways of instantiating this class: `EngineStatus()` and `EngineStatus.from_string()` that work differently. Should this go into `__init__` instead?

ptiurin (approved):

ptiurin:
We might need to add this to https://github.com/firebolt-db/firebolt-python-sdk/blob/main/src/firebolt/db/connection.py#L166 check as well

PR #367, opened 2024-03-28
fix: account_id handling for system engine in v2

ptiurin:
FIR-31838

Let backend add account_id to system engine when it's needed.

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -469,12 +469,15 @@ class CursorV2(Cursor):
                 self._set_parameters to be ignored.
         """
         parameters = parameters or {}
+        account_version = await self._client._account_version
         if use_set_parameters:
             parameters = {**(self._set_parameters or {}), **parameters}
         if self.parameters:
             parameters = {**self.parameters, **parameters}
-        # Engines v2 always require account_id
-        if self.connection._is_system or (await self._client._account_version) == 2:
+        # Engines v2 system engine will supply account_id dynamically in the url
+        if (self.connection._is_system and account_version == 1) or (
+            not self.connection._is_system and account_version == 2
+        ):
             assert isinstance(self._client, AsyncClientV2)
             parameters["account_id"] = await self._client.account_id
         return await self._client.request(
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -474,10 +474,8 @@ class CursorV2(Cursor):
             parameters = {**(self._set_parameters or {}), **parameters}
         if self.parameters:
             parameters = {**self.parameters, **parameters}
-        # Engines v2 system engine will supply account_id dynamically in the url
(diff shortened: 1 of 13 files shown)


ptiurin:
Integration tests succeeded https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/8469265115

src/firebolt/async_db/cursor.py
```diff
@@ -469,12 +469,15 @@ async def _api_request(
...
         parameters = parameters or {}
+        account_version = await self._client._account_version
         if use_set_parameters:
             parameters = {**(self._set_parameters or {}), **parameters}
         if self.parameters:
             parameters = {**self.parameters, **parameters}
-        # Engines v2 always require account_id
-        if self.connection._is_system or (await self._client._account_version) == 2:
+        # Engines v2 system engine will supply account_id dynamically in the url
+        if (self.connection._is_system and account_version == 1) or (
```
stepansergeevitch, line 478: Why do we have a `not self.connection._is_system and account_version == 2` branch? Does this mean we would add `account_id` to all user engines for v2 accounts?

stepansergeevitch: I believe this condition should just be `if self.connection._is_system and account_version == 1`

src/firebolt/utils/util.py
```diff
@@ -181,3 +190,21 @@ def __exit__(
...
             log_message = self._message + self.elapsed_time + "s"
             logger.debug(log_message)
+
+
+def parse_url_and_params(url: str) -> Tuple[str, Dict[str, str]]:
+    """Extract URL and query parameters separately from a URL."""
+    url = fix_url_schema(url)
+    parsed_url = urlparse(url)
+    query_params = parse_qs(parsed_url.query)
+    # This strips query parameters from the URL
```
stepansergeevitch, line 200: nit: Extend this comment a little. I was confused from the `urljoin` functionality first before I read the docs.

src/firebolt/utils/util.py
```diff
@@ -181,3 +190,21 @@ def __exit__(
...
+    result_url = urljoin(url, parsed_url.path)
+    # parse_qs returns a dictionary with values as lists.
+    # We want the last value in the list.
+    query_params_dict = {}
+    for key, values in query_params.items():
+        # Multiple values for the same key are not expected
+        if len(values) > 1:
+            raise ValueError(f"Multiple values found for key '{key}'")
+        query_params_dict[key] = values[-1]
```
stepansergeevitch, line 209: nit: values[-1] -> values[0], the same in this context but IMO more straightforward

src/firebolt/async_db/cursor.py
ptiurin: Do we not need to add `account_id` to user engines anymore? Reading https://packboard.atlassian.net/browse/FIR-31836 it specifies only system engine. If that's the case, I'm more than happy to remove this clause and simplify the expression.

stepansergeevitch (approved):

PR #369, opened 2024-04-10
fix: Use like in database fetch syntax

ptiurin:
split and any_match are no longer supported in Firebolt.
Working around it with LIKE syntax.

This code essentially searches for engine name in a text column. e.g. `my_test` in column value `some_engine (default), some_other_engine, my_test`

### DIFF
--- a/src/firebolt/service/V2/database.py
+++ b/src/firebolt/service/V2/database.py
@@ -85,16 +85,28 @@ class DatabaseService(BaseService):
                 condition.append("database_name like ?")
                 parameters.append(f"%{name_contains}%")
             if attached_engine_name_eq:
-                condition.append(
-                    "any_match(eng -> split_part(eng, ' ', 1) = ?,"
-                    " split(',', attached_engines))"
+                params = [
+                    # middle of the list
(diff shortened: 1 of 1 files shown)


alexradzin (commented):
How about to simplify the code as following. 
The query should wrap attached engines by commas and rewrite the full query as following:

```sql
SELECT 
    database_name, 
    concat(',', replace(replace(attached_engines, ', ', ','), ' (default)', ''), ',') as engins
FROM information_schema.databases
WHERE engines LIKE '%,?,%'
```
now the params array will look like 

```python
params = [attached_engine_name_eq]
```

Everything is hard coded, no dynamic query  creation.

stepansergeevitch:
Regarding @alexradzin suggested solution. Here `engines LIKE '%,?,%'` we seem to look for an engine name inside of the list, but it seems like it will ignore engines that are located on first and last positions of the list. Like in the list `engine1,engine2,engine3`, only engine2 will match. Am I missing something?

alexradzin:
> re located on first

no, please take take a look on `concat(',', ..., ',')`

PR #370, opened 2024-04-11
chore: Fir 30859 refactor out server side async execution support from python sdk

stepansergeevitch:
Removed all interface functions and logic for a deprecated server-side async execution support


src/firebolt/async_db/cursor.py
```diff
@@ -180,13 +177,7 @@ async def _do_execute(
...
             for query in queries:
                 start_time = time.time()
-                # Our CREATE EXTERNAL TABLE queries currently require credentials,
-                # so we will skip logging those queries.
-                # https://docs.firebolt.io/sql-reference/commands/create-external-table.html
-                if isinstance(query, SetParameter) or not re.search(
-                    "aws_key_id|credentials", query, flags=re.IGNORECASE
-                ):
-                    logger.debug(f"Running query: {query}")
+                self._log_query(query)
```
ptiurin, line 180: I think `_log_query` is defined as static so calling it with `self` might break things?

ptiurin (approved):

PR #371, opened 2024-04-11
test: Skip databases get_many

ptiurin:
This test is temporarily skipped since the underlying function is disabled in engines v1. Will revisit this later.

### DIFF
--- a/tests/integration/resource_manager/V2/test_database.py
+++ b/tests/integration/resource_manager/V2/test_database.py
@@ -1,3 +1,5 @@
+from pytest import mark
+
 from firebolt.client.auth import Auth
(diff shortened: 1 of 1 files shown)


tests/integration/resource_manager/V2/test_database.py
```diff
@@ -33,6 +35,7 @@ def test_database_get_default_engine(
     ], "Returned default engine name is neither of known engines"
 
 
+@mark.skip("split is not supported in engines v1 right now")
```
stepansergeevitch, line 38: Please add a task reference like `FIR-...` in a skip message

stepansergeevitch (approved):

PR #375, opened 2024-04-25
fix: Readd get_many for old engines

ptiurin:
get_many is missing since the version merge. Adding it as there might still be some use cases that rely on this.
Docs also mention this function.

Setting mypy to < 1.10.0 since the new version introduces errors outside of the scope of this change.


ptiurin:
V1 tests passing https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/8831167989

PR #376, opened 2024-04-25
ci: Decrease engine name length

ptiurin:
During nightlies some of the engine names became to long (>61 characters). This change avoids appending too much at the end.
e.g. https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/8826365388/job/24231982381


ptiurin:
Integration tests generally succeed, but fail due to a known bug.

stepansergeevitch (approved):

PR #378, opened 2024-05-06
fix: Engine running check in RM

ptiurin:
In order to execute engine start/stop commands we no longer need to verify the engine we're running is attached to a database since we're running against system engine.


ptiurin:
Merging since this is affecting a customer.

PR #383, opened 2024-05-22
test: Integration testing account v2

stepansergeevitch:
Created a separate integration testing workflow to run a full integration test suite on account v2. Made some fixes and improvements (mainly to resource management) to make sure tests are passing

### DIFF
--- a/src/firebolt/model/V2/engine.py
+++ b/src/firebolt/model/V2/engine.py
@@ -65,6 +65,7 @@ class Engine(FireboltBaseModel):
         "TYPE",
         "AUTO_STOP",
         "RENAME_TO",
+        # Empty strings for WARMUP and ENGINE_TYPE, since they are not supported in V2
         "",
         "",
     )
--- a/src/firebolt/model/V2/engine.py
+++ b/src/firebolt/model/V2/engine.py
@@ -74,7 +74,7 @@ class Engine(FireboltBaseModel):
 
     name: str = field(metadata={"db_name": "engine_name"})
     region: str = field()
-    spec: InstanceType = field(metadata={"db_name": "type"})
+    spec: Union[str, InstanceType] = field(metadata={"db_name": "type"})
     scale: int = field(metadata={"db_name": "nodes"})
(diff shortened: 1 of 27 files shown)


src/firebolt/model/V2/engine.py
```diff
@@ -60,14 +60,22 @@ class Engine(FireboltBaseModel):
...
         "ENGINE_TYPE",
     )
-    DROP_SQL: ClassVar[str] = "DROP ENGINE {}"
+    ALTER_PARAMETER_NAMES_V2: ClassVar[Tuple] = (
+        "NODES",
+        "TYPE",
+        "AUTO_STOP",
+        "RENAME_TO",
+        "",
+        "",
```
ptiurin, line 69: Why do we have empty strings here?

src/firebolt/service/V2/engine.py
```diff
@@ -129,11 +164,35 @@ def _format_engine_parameter(
...
+                raise ValueError(
+                    f"Parameters {bad_parameters} are not supported for this account"
+                )
+
+    def _format_engine_attribute_sql(
+        self, param: str, value: Union[str, int, EngineType, InstanceType, WarmupMethod]
+    ) -> Tuple[str, List]:
+        if param == "TYPE":
+            return f"{param} = {self._format_engine_parameter(value)} ", []
+        return f"{param} = ? ", [self._format_engine_parameter(value)]
```
ptiurin, line 189: Maybe let's pop a comment on what this method returns. Otherwise it's not clear what the List is supposed to be.

ptiurin (commented):
some minor comments

stepansergeevitch:
Integration tests: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/9269779587

src/firebolt/model/V2/engine.py
ptiurin: Fair enough, not feeling too strongly about it. Maybe let's just add a comment that this is left for compatibility.

PR #391, opened 2024-06-28
test: Fix sa name to be under 63 chars

ptiurin:
This fix is needed for nightlies to successfully pass.


ptiurin:
merging this as this is test-only fix and it would allow us to run nightlies over the weekend.

PR #392, opened 2024-07-01
feat: Fir 33643 replace i s databases with i s catalogs in drivers for python sdk

stepansergeevitch:
Added logic to replace `I_S.databases` with `I_S.catalogs` in databases service

### DIFF
--- a/src/firebolt/service/V2/database.py
+++ b/src/firebolt/service/V2/database.py
@@ -36,7 +36,6 @@ class DatabaseService(BaseService):
         return (
             f"{self.catalog_name}_name",
             "description",
-            "region",
             "uncompressed_size",
             "compressed_size",
             "attached_engines",
--- a/src/firebolt/service/V2/database.py
+++ b/src/firebolt/service/V2/database.py
@@ -1,16 +1,23 @@
 import logging
-from functools import cached_property
-from typing import List, Optional, Tuple, Union
+from typing import List, Optional, Union
 
 from firebolt.model.V2.database import Database
 from firebolt.model.V2.engine import Engine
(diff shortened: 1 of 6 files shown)


src/firebolt/service/V2/database.py
```diff
@@ -33,9 +21,44 @@ class DatabaseService(BaseService):
...
+    @cached_property
+    def get_sql(self) -> str:
+        return (
+            f"SELECT {', '.join(self.db_fields)} "
+            f"FROM information_schema.{self.catalog_name}s"
+        )
+
+    @cached_property
+    def get_by_name_sql(self) -> str:
+        return self.get_sql + " WHERE database_name=?"
```
ptiurin, line 56: small point, but I think at this stage `@cached_property` is an overkill. All properties it calls are also cached

src/firebolt/service/V2/database.py
ptiurin, line 1: Heads-up, `name_contains` will work for databases only - https://github.com/firebolt-db/firebolt-python-sdk/pull/392/files#diff-06fa197748434a9e62e51624141efe78a40d0c927caecdf888386955cfc9f993R110-R112

tests/unit/service/test_engine.py
```diff
@@ -94,6 +94,7 @@ def test_attach_to_database(
     database_get_callback: Callable,
     attach_engine_to_db_callback: Callable,
     system_engine_no_db_query_url: str,
+    select_one_query_callback: Callable,
```
ptiurin, line 97: we can probably get rid of this one if it's not used.

ptiurin (approved):
Otherwise lgtm.

PR #399, opened 2024-08-13
fix: Fir 35673 python sdk engine update fix renaming

stepansergeevitch:
Fixed engine renaming for fb2.0


tests/integration/resource_manager/V2/test_engine.py
```diff
@@ -139,3 +139,31 @@ def test_engine_update_multiple_parameters(
...
+    api_endpoint: str,
+    database_name: str,
+    engine_name: str,
+):
+    rm = ResourceManager(
+        auth=auth, account_name=account_name, api_endpoint=api_endpoint
+    )
+    name = f"{engine_name}_to_rename"
+    new_name = f"{engine_name}_renamed"
+    engine = rm.engines.create(name=name)
```
ptiurin, line 156: Ideally this would not start engine and create a stopped one so the whole test is faster and easier on the backend. However, looks like we don't expose autostart parameter, right?

src/firebolt/model/V2/engine.py
```diff
@@ -203,21 +208,31 @@ def update(
...
-        for param, value in zip(
-            self.ALTER_PARAMETER_NAMES,
-            (scale, spec, auto_stop, name),
-        ):
-            if value is not None:
-                sql_part, new_params = self._service._format_engine_attribute_sql(
-                    param, value
+        parameters: List[Union[str, int]] = []
+        if name is not None:
+            if not self._engine_name_re.match(name):
```
ptiurin, line 213: I'm thinking from a readability perspective, we can potentially do `name.replace('_', '').isalnum()` here and avoid regex altogether.
Don't feel too strongly about it, so up to you.

src/firebolt/model/V2/engine.py
```diff
@@ -194,6 +196,9 @@ def update(
             # Nothing to be updated
             return self
 
+        if name is not None and any(x is not None for x in (scale, spec, auto_stop)):
```
ptiurin, line 199: Would `if name and (scale or spec or auto_stop)` work here? Or do we want to explicitly allow empty strings?

src/firebolt/model/V2/engine.py
```diff
@@ -26,15 +27,16 @@ class Engine(FireboltBaseModel):
...
+    ALTER_PREFIX_SQL: ClassVar[str] = 'ALTER ENGINE "{}" '
     ALTER_PARAMETER_NAMES: ClassVar[Tuple] = (
         "NODES",
         "TYPE",
         "AUTO_STOP",
-        "RENAME_TO",
     )
     DROP_SQL: ClassVar[str] = 'DROP ENGINE "{}"'
 
+    _engine_name_re = re.compile(r"^[a-zA-Z0-9_]+$")
```
ptiurin, line 38: If we keep the regex approach let's pop a comment above it - "engine names can only be alphanumeric with underscores". Saves a second having to read the regex itself.

ptiurin (commented):
Some minor comments, otherwise looks good.

src/firebolt/model/V2/engine.py
stepansergeevitch: I feel like we should explicitly only consider a parameter not set in case it's None. In case it's empty, it might be an issue on a user end, they might have an empty string in the parameter value and pass it, and it will not get updated which will be confusing. An in current case, they will get an error for the empty string

tests/integration/resource_manager/V2/test_engine.py
stepansergeevitch: Correct, we don't expose that yet

src/firebolt/model/V2/engine.py
stepansergeevitch: I would rather stick to the regexp as it fells more straightforward, while removing characters might be slightly confusing during reading

ptiurin (approved):

PR #401, opened 2024-08-22
refactor: FIR-35468 remove account resolve endpoint from python sdk

stepansergeevitch:
- Removed usage of v2 `account/<account>/resolve` endpoint
- Removed usages v2 account id and account version

### DIFF
--- a/tests/unit/V1/client/conftest.py
+++ b/tests/unit/V1/client/conftest.py
@@ -7,7 +7,7 @@ from httpx import Request
 from pytest import fixture
 
 from firebolt.utils.exception import AccountNotFoundError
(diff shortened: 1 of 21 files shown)


tests/unit/V1/client/conftest.py
```diff
@@ -91,7 +91,7 @@ def check_credentials(
 
 @fixture
 def account_id_url(api_endpoint: str) -> Pattern:
-    base = f"https://{api_endpoint}{ACCOUNT_BY_NAME_URL}?account_name="
```
ptiurin, line 94: How did this work before? 😱

ptiurin (approved):
LGTM

tests/unit/V1/client/conftest.py
stepansergeevitch: Probably some fixture magic😊

PR #404, opened 2024-12-02
feat: FIR-38073 ecosystem support for geography type for python sdk

stepansergeevitch:
- Added support for geography type
- Simplified errors testing in integration tests

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,7 +55,7 @@ dev =
     devtools==0.7.0
     mypy==1.*,<1.10.0
     pre-commit==3.5.0
-    pyfakefs>=4.5.3
+    pyfakefs>=4.5.3,<=5.6.0
     pytest==7.2.0
     pytest-cov==3.0.0
     pytest-httpx>=0.13.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,8 @@ project_urls =
 packages = find:
 install_requires =
     aiorwlock==1.1.0
-    anyio>=3.7.1
+;   TODO: FIR-38687 Fix support for anyio 4.5.0+
+    anyio>=3.7.1,<4.5.0
     appdirs>=1.4.4
     appdirs-stubs>=0.1.0
     async-generator>=1.10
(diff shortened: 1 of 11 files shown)


tests/integration/dbapi/async/V2/test_errors_async.py
```diff
@@ -94,31 +85,9 @@ async def test_database_not_exists(
...
-    assert f"Database '{new_db_name}' does not exist or not authorized" in str(
-        exc_info.value
-    ), "Invalid database error message."
-
 
 async def test_sql_error(connection: Connection) -> None:
     """Connection properly reacts to SQL execution error."""
     with connection.cursor() as c:
-        with raises(OperationalError) as exc_info:
+        with raises(FireboltStructuredError):
```
ptiurin, line 92: Since structured error is such a broad error class, let's also test part of the error message here. And possibly other places.

stepansergeevitch: I don't like testing error messages because they change really often. That's why I've removed them. When we'll have error implemented property (with error codes), we'll be able to define error class by the error code and test this. For now I think it's ok to just test for the error presence

ptiurin: I agree, but not checking anything is not a solution here. We're checking that any error is thrown now - it could be auth error, network error, syntax, etc. We have to go with the lesser of two evils in here, which is IMO the error message check. We don't have to check the whole message, but at least a presence of a descriptive keyword. e.g. "parse" for parsing errors, etc.

ptiurin (approved):
LGTM, looking into why CI is failing

setup.cfg
```diff
@@ -26,7 +26,7 @@ project_urls =
 packages = find:
 install_requires =
     aiorwlock==1.1.0
-    anyio>=3.7.1
+    anyio>=3.7.1,<4.5.0
```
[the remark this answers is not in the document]
mistercrunch: nevermind, found this https://github.com/firebolt-db/firebolt-python-sdk/pull/404/files#r2356268086 !

PR #410, opened 2025-01-08
feat: FIR-42242 allow setting query timeout in python sdk

stepansergeevitch:
Added timeout_seconds parameter to cursor.execute.

### DIFF
--- a/src/firebolt/db/cursor.py
+++ b/src/firebolt/db/cursor.py
@@ -136,6 +136,7 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
             use_set_parameters: Optional[bool]: Some queries will fail if additional
                 set parameters are sent. Setting this to False will allow
                 self._set_parameters to be ignored.
+            timeout (Optional[float]): Request execution timeout in seconds
         """
         parameters = parameters or {}
         if use_set_parameters:
--- a/src/firebolt/db/cursor.py
+++ b/src/firebolt/db/cursor.py
@@ -202,7 +202,7 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
             for query in queries:
                 start_time = time.time()
                 Cursor._log_query(query)
-                timeout_controller.check_timeout()
+                timeout_controller.raise_if_timeout()
 
                 if isinstance(query, SetParameter):
                     row_set: RowSet = (-1, None, None, None)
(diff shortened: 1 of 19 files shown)


src/firebolt/db/cursor.py
```diff
@@ -108,20 +114,52 @@ def _raise_if_error(self, resp: Response) -> None:
...
         resp.raise_for_status()
 
-    @abstractmethod
     def _api_request(
         self,
         query: str = "",
         parameters: Optional[dict[str, Any]] = None,
         path: str = "",
         use_set_parameters: bool = True,
+        timeout: Optional[float] = None,
```
ptiurin, line 123: let's add this parameter to the docstring as well

src/firebolt/utils/timeout_controller.py
```diff
@@ -0,0 +1,29 @@
...
+    Used to check if the timeout has been reached and to calculate the remaining time.
+    """
+
+    timeout_error_cls = QueryTimeoutError
+
+    def __init__(self, timeout: Optional[float]):
+        self.timeout = timeout
+        self.start_time = time()
+
+    def check_timeout(self) -> None:
```
ptiurin, line 18: nit: `raise_if_timeout` name is more fitting here as it's clearer that this raises an error and not just checks if we've reached the correct time.

tests/integration/dbapi/async/V2/test_timeout.py
```diff
@@ -0,0 +1,12 @@
+from pytest import raises
+
+from firebolt.async_db import Connection
+from firebolt.utils.exception import QueryTimeoutError
+
+from .test_queries_async import LONG_SELECT
```
ptiurin, line 6: Let's use either relative or absolute imports like in this tests's [sync](https://github.com/firebolt-db/firebolt-python-sdk/pull/410/files#diff-5331bb2e13e9798e8f79a7f028be771940804514f551849da90c7313c1a7cbb6R1) counterpart.

ptiurin (commented):
Some small comments, overall lgtm.
Let's chat tomorrow on timeouts for multistatement.

tests/integration/dbapi/async/V2/test_timeout.py
stepansergeevitch: The problem here is that having a word `async` in import is treated as keyword and not as a module name. That's why I made id relative here.

stepansergeevitch: Made it relative in sync as well

ptiurin (approved):
As discussed, lgtm

PR #416, opened 2025-02-03
fix: escape single quotes with double single quotes

lucianosrp:
Escaping single quotes with a backslash  isn't working in Firebolt db v2 
This PR updates the single quote parsing with  two adjacent single quotes which is also accepted in v1. 

```select 'John\'s'``` <- works on v1 but not on v2
```select 'John''s'``` <- works on both


ptiurin:
Thanks for your PR @lucianosrp and well done on spotting this! I've created another [PR](https://github.com/firebolt-db/firebolt-python-sdk/pull/418) incorporating your changes as well as adding support for another case that was overlooked - `\`, which no longer needs to be escaped. Will wait for a review and aim to merge tomorrow!

lucianosrp:
Thanks for the speedy response on this! We really appreciate it as this is blocking one of our workflow.

ptiurin:
Closing this PR as the change was merged in #418 and released as a part of firebolt-sdk `1.9.1`

PR #417, opened 2025-02-03
fix(FIR-43199): add query cancel setting

ptiurin:
In production `cancel_query_on_connection_drop` setting by default does not cancel the query on lost network connection. Setting it explicitly in order to fix timeout behaviour.


ptiurin:
Moving back to draft as discussions if this is needed are still ongoing.

ptiurin:
No longer relevant, closing

PR #418, opened 2025-02-03
fix(FIR-43230): string escape in query parameters

ptiurin:
Build up upon https://github.com/firebolt-db/firebolt-python-sdk/pull/416
In Firebolt 2.0 we need to escape `'` with another quote - `''`. We no longer need to escape `\`.
However, in 1.0 slashes still need to be escaped.


src/firebolt/common/_types.py
```diff
@@ -372,158 +372,172 @@ def parse_value(
...
-        len(tokens) > 0
-        and tokens[0].ttype == TokenType.Keyword
-        and tokens[0].value.lower() == "set"
-    ):
-        # Check if set statement has a valid format
-        if len(tokens) == 2 and isinstance(tokens[1], Comparison):
-            return SetParameter(
-                statement_to_sql(tokens[1].left),
-                statement_to_sql(tokens[1].right).strip("'"),
+class StatementFormatter:
```
stepansergeevitch, line 389: I feel it makes sense now to give this class a personal file

src/firebolt/common/base_cursor.py
```diff
@@ -196,13 +198,16 @@ class BaseCursor:
...
+    def __init__(
+        self, *args: Any, escape_chars: Dict[str, str] = {}, **kwargs: Any
+    ) -> None:
         self._arraysize = self.default_arraysize
         # These fields initialized here for type annotations purpose
         self._rows: Optional[List[List[RawColType]]] = None
         self._descriptions: Optional[List[Column]] = None
         self._statistics: Optional[Statistics] = None
         self._row_sets: List[RowSet] = []
+        self._formatter = StatementFormatter(escape_chars)
```
stepansergeevitch, line 210: Let's pass formatter directly to constructor instead of `escape_chars`. escape_chars is a part of internal Formatter behavior, and should not be exposed.
Makes sense to even incapsulate contruction of formatter_v1 and formatter_v2 into functions somewhere near the Formatter class definition
Though it's just a nit

stepansergeevitch (approved):
Overall LGTM

PR #419, opened 2025-02-05
docs: remove docs repetition

elad-aharon:
cursor.fetchall() example was written twice in docs


ptiurin (approved):
Well spotted, thanks for the PR!

PR #423, opened 2025-04-08
refactor: FIR-43722 refactor our result set logic from cursors

stepansergeevitch:
Refactored out result storing and iteration into a separate RowSet class family. This is a prerequisite to implement support for streaming functionality

### DIFF
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -458,7 +458,7 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
     @check_not_closed
     def __enter__(self) -> Cursor:
         warnings.warn(
-            "Using __enter__ is deprecated, use async with instead", DeprecationWarning
+            "Using __enter__ is deprecated, use 'async with' instead", DeprecationWarning
         )
         return self
 
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 import time
+import warnings
 from abc import ABCMeta, abstractmethod
 from functools import wraps
 from types import TracebackType
@@ -383,6 +384,8 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
         assert self._row_set is not None
         with Timer(self._performance_log_message):
             # anext() is only supported in Python 3.10+
+            # this means we cannot just do return anext(self._row_set),
+            # we need to handle iteration manually
             try:
                 return await self._row_set.__anext__()
             except StopAsyncIteration:
@@ -451,6 +454,18 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
     async def __anext__(self) -> List[ColType]:
         assert self._row_set is not None
         return await self._row_set.__anext__()
+
+    @check_not_closed
+    def __enter__(self) -> Cursor:
+        warnings.warn(
+            "Using __enter__ is deprecated, use async with instead", DeprecationWarning
+        )
+        return self
+
+    def __exit__(
+        self, exc_type: type, exc_val: Exception, exc_tb: TracebackType
+    ) -> None:
+        return None
 
 
 class CursorV2(Cursor):
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -458,7 +458,8 @@ class Cursor(BaseCursor, metaclass=ABCMeta):
     @check_not_closed
     def __enter__(self) -> Cursor:
         warnings.warn(
-            "Using __enter__ is deprecated, use 'async with' instead", DeprecationWarning
+            "Using __enter__ is deprecated, use 'async with' instead",
+            DeprecationWarning,
         )
         return self
 
--- a/src/firebolt/async_db/cursor.py
+++ b/src/firebolt/async_db/cursor.py
@@ -5,16 +5,7 @@ import time
 from abc import ABCMeta, abstractmethod
 from functools import wraps
 from types import TracebackType
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Dict,
-    Iterator,
-    List,
-    Optional,
-    Sequence,
-    Union,
-)
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union
 from urllib.parse import urljoin
 
 from httpx import (
@@ -28,15 +19,6 @@ from httpx import (
 
 from firebolt.client.client import AsyncClient, AsyncClientV1, AsyncClientV2
 from firebolt.common._types import ColType, ParameterType, SetParameter
-from firebolt.common.base_cursor import (
-    BaseCursor,
-    _parse_update_endpoint,
-    _parse_update_parameters,
-    _raise_if_internal_set_parameter,
-    async_not_allowed,
-    check_not_closed,
-    check_query_executed,
-)
 from firebolt.common.constants import (
     JSON_OUTPUT_FORMAT,
     RESET_SESSION_HEADER,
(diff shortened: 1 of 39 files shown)


src/firebolt/async_db/cursor.py
```diff
@@ -121,13 +124,14 @@ async def _api_request(
...
         try:
-            return await self._client.request(
+            req = self._client.build_request(
                 url=urljoin(self.engine_url.rstrip("/") + "/", path or ""),
                 method="POST",
                 params=parameters,
                 content=query,
                 timeout=timeout if timeout is not None else USE_CLIENT_DEFAULT,
             )
+            return await self._client.send(req, stream=True)
```
goprean, line 134: was this the default, previous behavior with stream? I thought this is the refactoring and then you will add the streaming functionality.

goprean (commented):
I tried looking to see if I understand and can review the code but I don't know python so it is hard to review a change this big.

One note though: do you want to change the quality gate for python sdk to fail if there are new issues (like new violations or test coverage is not enough). I did change it for JDBC and now the email tells you that he quality gate failed. for python it passes even though there are new issues.

src/firebolt/async_db/cursor.py
stepansergeevitch: We used to read the response here fully and then work with the result in memory. Now we stream the response here, and let the RowSet to decide whether to read the stream in memory and process it or (in future streaming implementation) to read it line by line and not store it in memory

src/firebolt/common/row_set/types.py
```diff
@@ -0,0 +1,79 @@
...
+
+@dataclass
+class Column:
+    name: str
+    type_code: Union[type, ExtendedType]
+    display_size: Optional[int] = None
+    internal_size: Optional[int] = None
+    precision: Optional[int] = None
+    scale: Optional[int] = None
+    null_ok: Optional[bool] = None
```
ptiurin, line 63: This migration to dataclass breaks SQLAlchemy. It expects to be able to access column elements with `[]`.
e.g.
```
self = <sqlalchemy.engine.cursor.CursorResultMetaData object at 0x10551be60>
context = <sqlalchemy.engine.default.DefaultExecutionContext object at 0x105288110>
cursor_description = [Column(name='exists_', type_code=<class 'bool'>, display_size=None, internal_size=None, precision=None, scale=None, null_ok=None)]
    def _colnames_from_description(self, context, cursor_description):
        """Extract column names and data types from a cursor.description.
    
        Applies unicode decoding, column translation, "normalization",
        and case sensitivity rules to the names based on the dialect.
    
        """
    
        dialect = context.dialect
        translate_colname = context._translate_colname
        normalize_name = (
            dialect.normalize_name if dialect.requires_name_normalize else None
        )
        untranslated = None
    
        self._keys = []
    
        for idx, rec in enumerate(cursor_description):
>           colname = rec[0]
E           TypeError: 'Column' object is not subscriptable
venv/lib/python3.11/site-packages/sqlalchemy/engine/cursor.py:666: TypeError
```
Though I think there should be a way to ensure we're able to use square brackets here.

src/firebolt/async_db/cursor.py
```diff
@@ -353,75 +359,98 @@ async def executemany(
...
 
-    # TODO: figure out how to implement __aenter__ and __await__
     @check_not_closed
-    def __aenter__(self) -> Cursor:
+    async def __aenter__(self) -> Cursor:
         return self
 
-    @check_not_closed
-    def __enter__(self) -> Cursor:
-        return self
```
ptiurin, line 402: I think we can still support __enter__, at least to make sure we don't break anything. It's "free" from the looks of it.

src/firebolt/common/cursor/decorators.py
```diff
@@ -0,0 +1,59 @@
...
+    calling cursor method.
+    """
+
+    @wraps(func)
+    def inner(self: BaseCursor, *args: Any, **kwargs: Any) -> Any:
+        if self._state == CursorState.NONE or self._row_set is None:
+            raise QueryNotRunError(method_name=func.__name__)
+        if self._query_token:
+            # query_token is set only for async queries
+            raise MethodNotAllowedInAsyncError(method_name=func.__name__)
```
ptiurin, line 41: Do you think we can remove this part? Looks like we already have `async_not_allowed` decorator in all the places that this method is used.

src/firebolt/async_db/cursor.py
```diff
@@ -170,6 +174,9 @@ async def _validate_set_parameter(
         # set parameter passed validation
         self._set_parameters[parameter.name] = parameter.value
 
+        # append empty result set
+        await self._append_row_set_from_response(None)
+
```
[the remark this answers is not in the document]
stepansergeevitch: This has been the expected behavior most of the time. I'm not sure why and when we've stopped doing that. SET statements should behave as regular statements, and this will become inevitable when we'll switch to server-side set statements.

src/firebolt/async_db/cursor.py
```diff
@@ -353,75 +359,98 @@ async def executemany(
...
     async def fetchone(self) -> Optional[List[ColType]]:
         """Fetch the next row of a query result set."""
-        return super().fetchone()
+        assert self._row_set is not None
+        with Timer(self._performance_log_message):
+            # anext() is only supported in Python 3.10+
+            try:
+                return await self._row_set.__anext__()
+            except StopAsyncIteration:
+                return None
```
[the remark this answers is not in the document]
stepansergeevitch: Not sure I'm following. We're doing the whole `try:...catch:...` block instead of `anext(...)` just for backward compatibility. This works just fine with < 3.10

src/firebolt/common/row_set/types.py
stepansergeevitch: Hm, good catch. I'll upgrade it to have allow indexing.

src/firebolt/common/cursor/decorators.py
stepansergeevitch: Sure, I'll remove

src/firebolt/utils/async_util.py
```diff
@@ -0,0 +1,37 @@
...
+    result = []
+    try:
+        for _ in range(n):
+            result.append(await async_iterator.__anext__())
+    except StopAsyncIteration:
+        pass
+    return result
+
+
+def async_byte_stream(b: bytes) -> AsyncByteStream:
```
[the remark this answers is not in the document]
stepansergeevitch: Thanks for catching, it's outdated, will remove

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
ptiurin: I'm not sure I'm following the logic and comment here. So the comment says anext is not supported in Python 3.10 and below. So that means when we call `fetchone` we raise an error, since we're only catching StopAsyncIteration exception.

src/firebolt/async_db/cursor.py
ptiurin: But behind the scenes in `async aclose` we call `super.close()` so nothing stops us from having sync `close` pm cursor that would do the same. I believe one of the [reasons ](https://peps.python.org/pep-0492/#why-not-reuse-existing-magic-names) why `aenter` and `aexit` were introduced instead of having async `enter` and `exit` is to allow an object that's both sync and async context manager.

stepansergeevitch: After discussion, decided to leave a no-op function with a deprecation warning

src/firebolt/async_db/cursor.py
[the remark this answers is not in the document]
stepansergeevitch: Did some investigation, this seems to be a mistake we made during refactoring. Returning this functionality back as the desirable one

src/firebolt/async_db/cursor.py
```diff
@@ -353,75 +360,112 @@ async def executemany(
...
-        if row is None:
-            raise StopAsyncIteration
-        return row
+        assert self._row_set is not None
+        return await self._row_set.__anext__()
+
+    @check_not_closed
+    def __enter__(self) -> Cursor:
+        warnings.warn(
+            "Using __enter__ is deprecated, use async with instead", DeprecationWarning
```
ptiurin, line 461:
```suggestion
            "Using __enter__ is deprecated, use 'async with' instead", DeprecationWarning
```

ptiurin (approved):
Added suggestion for clarity, otherwise lgtm

PR #425, opened 2025-04-18
ci(NoTicket): Fix pre commit

ptiurin:
Pre-commit is using deprecated caching action and that action is no longer maintained. Instead a simple run command is used to run pre-commit checks.
Example [failure](https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/14528967276/job/40783760962).


goprean (approved):
If you are here do you want to also upgrade checkout to v4 and setup-python action to v4 as I have see that you have done for go.

PR #430, opened 2025-05-08
Expose _get_async_query_info

ymaayan:
It seems that `_get_async_query_info` can be really helpful for getting all the information about a running async query given a provided token.
Right now, it is only supported to either cancel a query or check if it finished or succeeded.
Getting more information about it like the `query_id` or `error_message` (if query failed) is necessary for easy monitoring/debugging purposes.

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -92,7 +92,7 @@ class Connection(BaseConnection):
         return c
 
     # Server-side async methods
-    async def _get_async_query_info(self, token: str) -> AsyncQueryInfo:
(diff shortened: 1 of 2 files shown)


ptiurin:
Thanks for bringing this up @ymaayan ! I've used this PR as a basis for #431 that has additional refactoring to make this method future proof as well as documentation. I'll close this PR in favour of #431 and will let you know once it's released.

ptiurin:
The aforementioned change is [released](https://github.com/firebolt-db/firebolt-python-sdk/releases/tag/1.12.0) in `firebolt-sdk==1.12.0`

PR #431, opened 2025-05-08
feat(FIR-45961): expose async query info method

ptiurin:
Initially proposed in https://github.com/firebolt-db/firebolt-python-sdk/pull/430

Exposing `get_async_query_info` method that allows users to easily access additional information about their async query: status, error message, query id, etc.

This includes refactoring to future-proof this method as multiple queries might correspond to a single async token.

### DIFF
--- a/src/firebolt/async_db/connection.py
+++ b/src/firebolt/async_db/connection.py
@@ -17,6 +17,7 @@ from firebolt.common.base_connection import (
     ASYNC_QUERY_STATUS_SUCCESSFUL,
     AsyncQueryInfo,
     BaseConnection,
+    _parse_async_query_info_results,
 )
 from firebolt.common.cache import _firebolt_system_engine_cache
 from firebolt.common.constants import DEFAULT_TIMEOUT_SECONDS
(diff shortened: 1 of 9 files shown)


src/firebolt/async_db/connection.py
```diff
@@ -92,34 +93,41 @@ def cursor(self, **kwargs: Any) -> Cursor:
...
-                "Rerun the command and reach out to Firebolt support if "
-                "the issue persists."
+    def _raise_if_multiple_async_results(
+        self, async_query_info: List[AsyncQueryInfo]
+    ) -> None:
+        # We expect only one result in current implementation
+        if len(async_query_info) != 1:
+            raise NotImplementedError(
+                "Async query status request returned more than one result. "
+                "This is not supported yet."
```
goprean, line 128: If this happens then it seems it is a backend issue, right? What should the client do with this error message? Should we mention to contact Firebolt support. This does not seem like a client issue.

goprean (approved):
lgtm.
Can you add a comment on the PR that was issued with just exposing the get_async_query_data so the developer knows what we are fixing his PR but in a different way.

src/firebolt/async_db/connection.py
ptiurin: I don't think support is able to do anything. This is future-proofing the implementation when several queries can be run in the same async "batch" but we haven't updated the driver. On the server side this is working as intended, but on our side it's not implemented - hence the `NotImplementedError`.

PR #433, opened 2025-05-21
docs: NoTicket fix dbapi examples notebook

stepansergeevitch:
Fixed dbapi examples notebook to work on GitHub preview


examples/dbapi.ipynb
```diff
@@ -410,17 +422,24 @@
    "cell_type": "raw",
```
ptiurin, line 422: let's change this to `markdown` while we at it. Otherwise it's different from other headings.

ptiurin (approved):
apart from that lgtm

PR #447, opened 2025-07-16
Update dependabot PR title generation

stepansergeevitch:
Update how dependabot generates PR titles to pass our PR title checks


ptiurin (approved):

goprean (approved):
Nice! I think we should also update all of the other repos with this prefix message. That way is consistent. I think for JDBC it is the default commit message.

PR #462, opened 2025-10-07
chore(NoTicket): Improve plaintext error handling

ptiurin:
We assumed all the errors that are not recognised by us will be sent in a structured format, with JSON.
However, doesn't look like it's always the case.
We should be able to gracefully handle plaintext errors as well.

### DIFF
--- a/src/firebolt/utils/util.py
+++ b/src/firebolt/utils/util.py
@@ -19,6 +19,7 @@ from httpx import URL, Response, codes
 
 from firebolt.utils.exception import (
     ConfigurationError,
+    FireboltError,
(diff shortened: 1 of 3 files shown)


goprean:
> > lgtm. Just for my education, what backend/3rd party call raises an error that is not a json?
> 
> The one I stumbled upon happens when engine is stopped and autostart is off. If you try to query it you get an error in plaintext.

In my opinion this should be fix on the backend. The same API returning different error formats depending on the state of the engine. Could you at least ask on packdb or gateway to see if this is the desired behavior? Because otherwise we would have to change other drivers as well.

PR #493, opened 2026-03-22
feat: add option to run core integration tests on kind

fstr:
**Problem**
1. The current firebolt-python-sdk integration test suite for core relies on docker-compose. The docker-compose setup is not part of the repository and is instead set up via a reusable GitHub Action. The compose file could be moved into this repository, but the more different cluster variations you want to test with the SDK, the more cumbersome the compose setup gets.
2. The compose setup is only run once before all integration tests are executed. If a test requires a different setup, the compose must be part of the test driver itself.

**Trade Off**
Both issues could be resolved while sticking to compose, but we already have a firebolt-core Helm chart published in GHCR, which makes multi-node cluster setups straight forward and undergoes continuous testing. If there are more components released via Helm charts in the future, they can easily be integrated. Kind has the downside that we need to work with `kubectl port-forward` to reach the core containers, so it's another layer in the test setup.

**Implementation**
- Add option to run Firebolt Core tests on a local kind cluster instead of using docker-compose with the goal of phasing out the docker-compose setup. For now, we can run the integration test suite with `--run-compose` or `--run-kind` or both. This approach was also chosen to show that there are no breaking changes.
- Update fixtures to depend on `kind_app_setup`, as the CORE_URL is only known after we installed the Helm chart and ran kubectl port-forward
- We set-up a kind cluster with a single Firebolt Core Helm chart installation at the beginning of the integration test suite. All tests run against this installation. This mirrors the docker-compose setup.
  - Add option to do additional customized Helm installations per test function/method. For example a 5 node cluster.

### DIFF
--- a/.github/workflows/integration-tests-core.yml
+++ b/.github/workflows/integration-tests-core.yml
@@ -69,7 +69,7 @@ jobs:
           python -m pip install --upgrade pip
           pip install ".[dev]"
 
-      - name: Run integration tests HTTP
+      - name: "Compose: Run integration tests HTTP"
         env:
           SERVICE_ID: ${{ secrets.FIREBOLT_CLIENT_ID_STG_NEW_IDN }}
           SERVICE_SECRET: ${{ secrets.FIREBOLT_CLIENT_SECRET_STG_NEW_IDN }}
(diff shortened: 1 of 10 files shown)


fstr:
Test run: https://github.com/firebolt-db/firebolt-python-sdk/actions/runs/23402812285/job/68076711457

gm42 (approved):

kop (dismissed):
Makes perfect sense!

gm42 (dismissed):

PR #509, opened 2026-04-14
fix: don't inherit query_label on validation queries (FB-626)

fstr:
**Description**
SET parameter validation runs a SELECT 1 probe. That probe no longer carries an inherited query_label from earlier SETs, so query history is not polluted by the validation calls when filtering on the query label.

- _api_request accepts optional drop_parameter_keys. When empty/omitted, behavior is unchanged (no extra work on the hot path).
- Non-empty: remove listed keys from merged URL params unless the key is in this call’s explicit parameters overlay.
- _validate_set_parameter passes SET_VALIDATION_DROP_PARAMETER_KEYS (query_label only)


gm42 (approved):

fstr:
Closing. The `SET query_label` call itself will still be labeled but then have inconsistent behavior compared to other SET calls.

